PR #92, opened 2015-09-01
[ready] E-book form conversion & validation

romkebon:
- [x] The user doesn't get to see the thanks page after downloading the e-book. Therefore the conversion is now tracked on form submission.
- [x] Check for e-mail address (for browsers that don't support `required` and `type="email"`).

### DIFF
--- a/source/javascripts/_ebook-form.js
+++ b/source/javascripts/_ebook-form.js
@@ -10,6 +10,13 @@ Defacto.ebookForm = {
     event.preventDefault();
     $submit.prop('disabled', 'disabled');
 
+    //Validate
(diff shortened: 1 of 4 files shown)


source/javascripts/_ebook-form.js
```diff
@@ -51,6 +52,11 @@ $(function () {
     // Download ebook
     window.open(ebookUrl, '_blank');
 
+    // Track coversion
+    if(ga) {
```
fatboypunk: could use a space :dog:

sn3p:
nice, much cleaner :+1:

PR #130, opened 2015-12-21
[READY] Optimize the Workshop landing page

RonaldDijkstra:
- [x] Add Schema.org Event markup for SEO purposes
- [x] Optimize content (textual)
- [x] Add sharing options 
- [x] Layout fixes (conversion rate optimization)
- [x] some enters for readability
- [x] a more specific URL (e.g. /workshop-convenant-mt or /workshop-convenant-medische-technologie) 
- [x] Update of CTA's anchor text

PR #133, opened 2016-01-04
[Ready] Blog Joep about Convenant MT

RonaldDijkstra:
A blog about the Convenant and workshops.


RonaldDijkstra:
ready for review

source/workshop-convenant-mt.html.erb
```diff
@@ -0,0 +1,158 @@
+---
+title: "Workshop: Hoe borg ik het Convenant MT? - 28 januari 2016"
```
sn3p: Does this belong here? Looks like this is in conflict with https://github.com/DefactoSoftware/website/pull/130

RonaldDijkstra: True, this a mistake

Issue #151, opened 2016-01-15
Width LS hero logo on LS productpage on mobile


RonaldDijkstra:
it doesn't fit on mobile

Tested Chrome + safari 

http://www.defacto.nl/learningspaces
http://www.defacto.nl/de/learningspaces

RonaldDijkstra:
![learningspaces defacto](https://cloud.githubusercontent.com/assets/16101007/12349896/64b55d22-bb73-11e5-92c1-d32d6b9ccdce.png)

PR #153, opened 2016-01-22
[abandoned] Ronald/content and layout improvements

RonaldDijkstra:
PR for the changes I'm working on. 

- [ ] Remove featured blog from homepage (see comments)
- [x] Only one testimonial on homepage and a button to more testimonials
- [ ] Set just one macro conversion goal per page
- [ ] add LMS to CAPP titles 
- [ ] improve content on homepage, producten, diensten, kwaliteitspaspoort, learningspaces, over ons (Rob), contact (Rob)
- [ ] new page /capp-leren (we don't know where this is placed in the site structure yet)
- [ ] new page /capp-ontwikkelen (same)
- [ ] set better URL's for capp, elearning, hosting (for ex. hosting page covers hosting AND security, but there's no mention of that in the URL)

### DIFF
--- a/source/localizable/index.html.erb
+++ b/source/localizable/index.html.erb
@@ -22,18 +22,21 @@ sitemap_priority: 1
 
 <%= partial "product_highlights" %>
 
-<section class="testimonials cols">
+<section class="testimonials">
   <div class="headline">
     <h2><%= t("testimonials.title") %></h2>
   </div>
   <div class="row">
 
     <%= partial "testimonials", locals: {
-      count: 2,
+      count: 1,
       featured: true,
       summary: true
     } %>
 
+  </div>
+  <div class="row">
+    <%= nav_link_to t("testimonials.more_testimonials"), "/testimonials.html", class: "button blue outline" %>
   </div>
 </section>
 
--- a/source/localizable/index.html.erb
+++ b/source/localizable/index.html.erb
@@ -37,5 +37,5 @@ sitemap_priority: 1
   </div>
 </section>
 
-<%= partial "blog/featured" %>
+<%#= partial "blog/featured" %>
 <%= partial "clients" %>
--- a/source/localizable/learningspaces.html.erb
+++ b/source/localizable/learningspaces.html.erb
@@ -5,8 +5,8 @@ title: LearningSpaces
 <div class="hero">
   <div class="hero-inner">
     <div class="hero-copy">
-      <h1><%= image_tag "logos/learningspaces-white.png", alt: "LearningSpaces" %></h1>
-      <%= markitdown t("learningspaces.hero.text_md") %>
+      <%= image_tag "logos/learningspaces-white.png", alt: "LearningSpaces" %>
+      <h1><%= t("learningspaces.hero.text_md") %></h1>
     </div>
 
     <% if I18n.locale == :nl %>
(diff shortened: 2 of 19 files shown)


RonaldDijkstra:
0a696e2 fixes #151

source/localizable/index.html.erb
```diff
@@ -22,20 +22,23 @@ sitemap_priority: 1
...
     } %>
 
   </div>
+  <div class="row">
+    <%= nav_link_to t("testimonials.more_testimonials"), "/testimonials.html", class: "button blue outline" %>
+  </div>
 </section>
 
-<%= partial "blog/featured" %>
+<%#= partial "blog/featured" %>
```
fatboypunk: why would we remove it?
If it should be removed you don't have to leave it as a comment

source/stylesheets/_learningspaces.scss
```diff
@@ -5,11 +5,19 @@
...
-      //   font-size: 4em;
-      // }
+      font-size: 2em;
+      font-weight: 300;
+      line-height: 1.2em;
+      margin-bottom: 0.75em;
+    
+      @include media($small-only) {
+      font-size: 1em;
+      margin-bottom: 1.5em;
```
fatboypunk: indentation :dog:

source/stylesheets/_producten.scss
```diff
@@ -1,7 +1,9 @@
 .producten_index,
 .diensten_index {
   text-align: center;
-
+  h2 {
```
fatboypunk: enter :dog:

source/stylesheets/_producten.scss
```diff
@@ -1,7 +1,9 @@
 .producten_index,
 .diensten_index {
   text-align: center;
-
+  h2 { 
+    font-size: 1.5em
+  }
   .solitions-list {
```
fatboypunk: enter :dog2:

source/localizable/learningspaces.html.erb
```diff
@@ -5,8 +5,8 @@ title: LearningSpaces
 <div class="hero">
   <div class="hero-inner">
     <div class="hero-copy">
-      <h1><%= image_tag "logos/learningspaces-white.png", alt: "LearningSpaces" %></h1>
-      <%= markitdown t("learningspaces.hero.text_md") %>
+      <%= image_tag "logos/learningspaces-white.png", alt: "LearningSpaces" %>
+      <h1><%= t("learningspaces.hero.text_md") %></h1>
```
fatboypunk: I don't think this is the way to fix the problem you were addressing. Also I feel like the h1 should be a header for the text, while here the h1 is the text.
Don't you lose the markup without the markdown helper now?

source/localizable/index.html.erb
RonaldDijkstra: There was a discussion about this on tuesday. RedMonkey said in the original document: 

"Niet de juiste plaats voor een blog, alle blogs onder het Blogmenu"

My opinion on this matter is that this a nice place for a featured blog: it doesn't hurt SEO and it's good to show we're relevant. There is an argument in the traffic too, because the number of entrances on the homepage are way higher than on the blog page. So if we want our content to be read, we should have a featured blog on the homepage.

source/localizable/learningspaces.html.erb
RonaldDijkstra: It does, will find a better solution. Thanks again.

source/localizable/index.html.erb
RonaldDijkstra: Looking at the data of past month we can assume this part of the homepage is not used that much. 

We should start tracking clicks on the more button.

PR #163, opened 2016-02-05
[READY]] Ronald/german video homepage

RonaldDijkstra:


RonaldDijkstra:
@sn3p will you check this? Thanks in advance!

sn3p:
Awesome @RonaldDijkstra :+1: 

I renamed the partials and modified the markup/breakpoints a bit for better mobile readability.
If you agree with this you may merge.

source/stylesheets/main.scss
```diff
@@ -33,6 +33,8 @@
 @import "workshop";
 @import "cta_workshop";
 @import "announcement";
+@import "german_news";
```
fatboypunk: Don't we want this to be just news? So that if we have dutch or english news we can use this as well?
Or is it to specific?

RonaldDijkstra: Nice suggestion, the news part could be something on the dutch website as well.

sn3p: Suppose we could, but right now we're only doing this for the German site:
```erb
<%= partial "german_news" if I18n.locale == :de  %>
```

PR #183, opened 2016-03-29
[WIP] set a local virtual pageview for conversions

RonaldDijkstra:
So we could distinguish the Dutch and German conversions. 

- [x] send 'danke' instead of 'bedankt' for german conversions
- [x]  setup tracking of /danke/ conversions in GA


fatboypunk:
needs a rebase

RonaldDijkstra:
Obsolete, since we have a .de domain

PR #185, opened 2016-04-08
[READY] proper way of domain linking with GA

RonaldDijkstra:
Added allowlinker:true


source/localizable/_google_analytics.erb
```diff
@@ -4,10 +4,10 @@
   m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
   })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
 
-  ga('create', 'UA-6700447-1', 'auto');
+  ga('create', 'UA-6700447-1', 'auto' {'allowLinker': true});
```
fatboypunk: misses a comma after auto

RonaldDijkstra: Oops..

PR #197, opened 2016-05-10
[READY] New homepage 

tarzan:
rebased @RonaldDijkstra 's work


source/stylesheets/_layout.scss
```diff
@@ -98,7 +98,7 @@ article {
     margin: 0 0 0.8em 0;
     padding: 0 0 0 30px;
 
-    &:before {
+    &::before {
```
RonaldDijkstra: We need all 12 properties here.

sn3p:
@RonaldDijkstra think we are good to go. Made some final changes, let me know what you think.

NL
---

defacto-nl

DE
---

defacto-de

RonaldDijkstra:
@sn3p we're good yes. I think your changes are awesome and the way we should go. Let's wait no longer...

PR #201, opened 2016-05-17
[READY] change de to duitsland on links to german jobs

RonaldDijkstra:
Minor change in text on the about us page


fatboypunk:
Why would we do this? Also it's only in dutch now

RonaldDijkstra:
Because Hans asked me to change (de) to (Duitsland)..

PR #204, opened 2016-05-17
[READY] team update

RonaldDijkstra:
#200


fatboypunk:
The other newly added pictures are 160x160 do we want these at the same widthxheight or is this the best we have? or other reasons? ^^

RonaldDijkstra:
Hm guess those where wrong then, because I checked in the source directory and found most pics to be 100x100.

fatboypunk:
Yeah but i thought we did the new ones bigger for retina results.

PR #216, opened 2016-05-24
[WIP] fix broken link on cmt page

RonaldDijkstra:
This is ready for review. 

On [this page](http://www.defacto.nl/convenant-medische-technologie/) the German flag points to a non existing page, so I pointed the flag to the German homepage. 

Maybe there could be a more sustainable way to do this?

### DIFF
--- a/source/localizable/_header.erb
+++ b/source/localizable/_header.erb
@@ -18,7 +18,8 @@
         <% end %>
         <li><%= nav_link_to t("header.contact"), "/contact.html" %></li>
         <li class="flags">
-          <%= country_flags %>
+          <%= country_flags unless current_page.url == "/convenant-medische-technologie/" %>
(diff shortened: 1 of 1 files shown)


source/localizable/_header.erb
```diff
@@ -18,7 +18,8 @@
         <% end %>
         <li><%= nav_link_to t("header.contact"), "/contact.html" %></li>
         <li class="flags">
-          <%= country_flags %>
+          <%= country_flags unless current_page.url == "/convenant-medische-technologie/" %>
```
fatboypunk: @sn3p maybe it's a good idea to make a method for this that can check multiple urls?

fatboypunk: Also we still want the country_flags right? just not point them to the current_page url?

RonaldDijkstra: In one of the visits from last weekend I saw a German user switching to the German domain.

RonaldDijkstra: Thnx for comments, there's another PR now with another approach to this problem.

PR #218, opened 2016-05-30
[READY] use hreflangs for locales

RonaldDijkstra:
Even last weekend we saw German users land on the NL TLD, so we need to make clear for search engines which version of the Defacto site is which. With a bit of help from @sn3p we made a helper to set hreflangs in each version of the site. 

Some more info on this: 

https://support.google.com/webmasters/answer/189077?hl=nl 
https://moz.com/learn/seo/hreflang-tag

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -261,6 +261,18 @@ helpers do
     html
   end
 
+  # Href langs
+  def href_langs
+    html = ""
+      langs.each do |lang|
+      locale_root_path = current_page.locale_root_path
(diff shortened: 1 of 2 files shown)


sn3p:
👍 @RonaldDijkstra looks good, just need to fix the inconsistent indentation (see hound)

RonaldDijkstra:
search console internationale targeting http www defactolearning de

Search Console always has a litte delay, but there's something going on now with the hreflangs. :+1:

PR #229, opened 2016-06-07
[READY] Flag and hreflang links to other language

RonaldDijkstra:
Checks if the page exists in another language. If not the flag should point to the homepage of the other language. 

- [x] flags
- [x] hreflangs
- [x] rebase

On Defactolearning.de it seems to only link to the homepage for now. 

See the switch here to see how it works when there's no page on the other website: http://www.defacto.nl/convenant-medische-technologie/

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -264,11 +264,12 @@ helpers do
       img = image_tag("flags/#{lang}.gif", alt: flag_titles[lang])
       if current_page.url == "/blog/"
         url = full_url("/blog", lang)
+      elsif current_page.url == "/convenant-medische-technologie/"
+        url = full_url("", lang)
       else
         locale_root_path = current_page.locale_root_path
         url = locale_root_path ? locale_root_path : "/"
       end
-      #url = sitemap.find_resource_by_path(url) ? url : "/"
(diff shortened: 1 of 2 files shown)


RonaldDijkstra:
Couldn't make the check in the sitemap work on Defactolearning.de, so instead it checks the current page, in config.rb for the flag links and in the source of layout for hreflangs. We need those removed too on the CMT page. 

This suits for now, because we only have 1 page on the dutch website that's not on the german site (besides blog articles). But maybe not in the future if both sites will get more different from eachother.

RonaldDijkstra:
Find resource by path stays empty on all instances when used on DefactoLearning.de. Maybe something with forcing the page classes to not translate?
http://rubydoc.info/gems/middleman-core/Middleman/Sitemap/Store#find_resource_by_path-instance_method

fatboypunk:
Also an option could be that we just make a list of pages that do not exist in the other language

Can you make an issue to solve this in a better way? so we won't forget to do it :) I'll check this PR tomorrow

RonaldDijkstra:
#239

fatboypunk:
> #239

<3

RonaldDijkstra:
And thanks

PR #257, opened 2016-07-12
[READY] Center blog posts again

RonaldDijkstra:


RonaldDijkstra:
The blog posts went far to the left 

With this: 

`body[class^="blog_"],
body[class*="_index"]`

The `.post-body` on the blog post got targeted. AFAIK we don't need this class on the blog post layout, as can be seen on the screenshot below: 
bevoegd is niet bekwaam waar bekwaamheid echt over gaat defacto

fatboypunk:
Wouldn't it be better to change the css blocks so bot are needed before it get's executed not just one?

fatboypunk:
I think just dropping the comma would be enough

RonaldDijkstra:
Then we're losing some ofthe style on the indexes. 

blog2 defacto

PR #259, opened 2016-07-15
add blog of marian

fatboypunk:


fatboypunk:
Where in the text do you want te picture to be placed? (if at all?) @IwannadeJonge ?

PR #283, opened 2016-07-27
[READY] Hound settings

RonaldDijkstra:
I'd like to make some changes to Hound settings, like the depths of nesting and selector depths. I made 2 commits, one with the default settings and one with changes. 

I have a few questions about the settings LS uses, becasue they differ from the default: 

LS: 
```
 ExtendDirective:
  enabled: true

QualifyingElement:
  enabled: true
  allow_element_with_attribute: true
  allow_element_with_class: true
  allow_element_with_id: false

SelectorFormat:
  enabled: true
  convention: hyphenated_BEM

SpaceAfterVariableColon:
  enabled: false
  style: one_space 
```

Default:
```
ExtendDirective:
  enabled: false

QualifyingElement:
  enabled: true
  allow_element_with_attribute: false
  allow_element_with_class: false
  allow_element_with_id: false

SelectorFormat:
  enabled: true
  convention: hyphenated_lowercase

SpaceAfterVariableColon:
  enabled: true
  style: one_space
```
Would like to talk tomorrow about these in a spare moment, because I'd want to know where they are for. @fatboypunk or @sn3p

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,16 @@
+# Hound config - https://houndci.com/configuration
+# ------------------------------------------------
+# We try to follow the Thoughtbot style guides:
+# https://github.com/thoughtbot/guides
+
+javascript:
+  enabled: false
+
+jscs:
+  enabled: false
+
+ruby:
+  enabled: true
+
+scss:
+  config_file: .scss-lint.yml
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -1,0 +1,253 @@
+# Default application configuration that all configurations inherit from.
+
+scss_files: "**/*.scss"
+plugin_directories: ['.scss-linters']
+
+# List of gem names to load custom linters from (make sure they are already
+# installed)
+plugin_gems: []
+
+# Default severity of all linters.
+severity: warning
+
+linters:
+  BangFormat:
+    enabled: true
+    space_before_bang: true
+    space_after_bang: false
+
+  BemDepth:
+    enabled: false
+    max_elements: 1
+
+  BorderZero:
+    enabled: true
+    convention: zero # or `none`
+
+  ChainedClasses:
+    enabled: false
+
+  ColorKeyword:
+    enabled: true
+
+  ColorVariable:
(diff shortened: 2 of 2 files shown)


fatboypunk:
https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md
As a reference

RonaldDijkstra:
Thanks for that, maybe that's enough

RonaldDijkstra:
Is enough to understand what every rule is for. 

1. This website doesn't use @extend, LS landing does so this could be true 
2. QualifyingElement should be true for classes, because we use `body.classname` like `body.bedankt` from now on
3. I can understand why you use **Hyphenated_BEM** because LS is a big project, but I guess the website doesn't need this, so hyphenated_lowercase should be it
4. And SpaceAfterVariableColon I think must be true for the website

sn3p:
> 1. This website doesn't use **@extend**, LS landing does so this could be true
2. **QualifyingElement** should be true for classes, because we use `body.classname` like `body.bedankt` from now on
3. I can understand why you use **Hyphenated_BEM** because LS is a big project, but I guess the website doesn't need this, so hyphenated_lowercase should be it
4. And **SpaceAfterVariableColon** I think must be true for the website

Nice @RonaldDijkstra 

1. Don't we use `@extend` for bourbon mixins?
2. Yep, bad practice in general but for us body classes is an exception imo.
3. I prefer `hyphenated_lowercase` but Middleman/rails generated `snake_case` classes. In LS we have **SelectorFormat** disabled for this same reason, so I'd do the same for the website.
4. Enable **SpaceAfterVariableColon** with style: `at_least_one_space`.

RonaldDijkstra:
> 1. Don't we use @extend for bourbon mixins?

AFAIK not: 

```
Searching 42 files for "@extend"

0 matches across 0 files
```

sn3p:
Ah doh, I confused `@extend` with `@include`.

RonaldDijkstra:
Thanks @sn3p

Issue #277, opened 2016-07-26
With build errors deploying is still possible


RonaldDijkstra:
- ~~Since yesterday building is really really slow (even time outs sometimes)~~
- If build errors deploying is still possible

fatboypunk:
Locally for me the about us page is very slow, might have something to do with that

sn3p:
> Locally for me the about us page is very slow, might have something to do with that

About page has always been slower because team (gr)avatars are being fetched.

fatboypunk:
> About page has always been slower because team (gr)avatars are being fetched.

good to know :)

RonaldDijkstra:
That page won't build, and some blogs won't build as wel.. They time-out. 

```
       error  build/blog/wat-we-hebben-geleerd-van-het-opensourcen-van-ons-hobbyproject/index.html
== Instrument (render.resource): 60025.92434799999ms
```

:(

RonaldDijkstra:
It's still slow but at least I can build now.

RonaldDijkstra:
Build problem is gone for now.

PR #285, opened 2016-07-28
[READY] Prevent deployment when build fails

sn3p:
Fixing https://github.com/DefactoSoftware/website/issues/277

The `rake deploy` task now exits when the the build fails.

screen shot 2016-07-28 at 11 59 45

screen shot 2016-07-28 at 12 00 38

This differs from the LearningSpaces landing page we're using a [`deploy` bash script](https://github.com/DefactoSoftware/LearningSpaces-Landing/blob/master/deploy).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -22,9 +22,9 @@ end
 namespace :build do
   def build(env)
     puts "*"*50
-    puts "* Building #{env.upcase}"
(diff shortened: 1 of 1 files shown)


RonaldDijkstra:
Thanks, successfully tested! 

successfully-test

PR #291, opened 2016-08-04
[READY] Localize last email

### LINKED ISSUE #273: @defactolearning.de email
het duitsland team will graag @defactolearning.de emails die dan ook in outlook terecht komen :)

RonaldDijkstra:
- localize email on `contact_info` too
- remove redundant phone numbers from locales
- use `unless` instead of `if`

Fixes #273

### DIFF
--- a/source/localizable/_footer.erb
+++ b/source/localizable/_footer.erb
@@ -44,7 +44,7 @@
       <p>
         Zuiderpark 3<br>
         9724 AD Groningen<br>
-        <%= "Niederlande" unless I18n.locale == :nl %>
+        <%= t("footer.country") %>
       </p><p>
         T&nbsp;&nbsp;<a href="tel:<%= t("contact.phone_link") %>" class="phone"><%= t("contact.phone") %></a><br>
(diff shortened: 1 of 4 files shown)


source/localizable/_footer.erb
```diff
@@ -44,12 +44,10 @@
       <p>
         Zuiderpark 3<br>
         9724 AD Groningen<br>
-        <% if I18n.locale == :de %>
-          Niederlande
-        <% end %>
+        <%= "Niederlande" unless I18n.locale == :nl %>
```
fatboypunk: I think an if would be better here, if the pages are going to be translated to english you'd have to change this as well

fatboypunk: Nice that you made it a oneliner though :D

RonaldDijkstra: @sn3p said a translated string would be best (empty in nl.yml).

fatboypunk: he's right :)

RonaldDijkstra: Ok

PR #308, opened 2016-09-27
[ready] change formal into informal

jwcoens:


fatboypunk:
It's the good old switcheroo

PR #336, opened 2016-10-06
[READY] Remove contact from about us sidebar

RonaldDijkstra:
Because it's already in the main nav I removed the contact link from the about us sidebar.


fatboypunk:
But then you can't return to that page without going to the top bar?

RonaldDijkstra:
Yup. Got this as a request. Keeping it there is also fine by me.

RonaldDijkstra:
If the sidebar menu stays on the contact page, the link should not be removed. Closed for now, thanks for pointing this out @fatboypunk

PR #340, opened 2016-10-12
[READY] Update KPP Page

RonaldDijkstra:
- [x] change hero BG color
- [x] update German content
- [x] add image to hero


source/stylesheets/main.scss
```diff
@@ -54,10 +70,10 @@ figcaption {
 // Hide Google Conversion iframe
 iframe[name="google_conversion_frame"] {
   display: none !important;
+  float: left;
+  font-size: 0 !important;
```
fatboypunk: We can't do without the importants?

RonaldDijkstra: I was checking that. We use the same as most approved here: http://stackoverflow.com/questions/18467529/google-remarketing-tag-iframe-height-issue

Tried to test it in older browsers, where apparently these importants are for. Definitely something I am looking into for a different PR somewhere in the near future!

Issue #239, opened 2016-06-28
Sustainable check if page exists on other TLD


RonaldDijkstra:
We need a more sustainable way of checking if the page has it's equivalent on the other TLD.

dkhgh:
Haven't looked at this in a while, wasn't it the case that **every** `page_name.erb` ~~is~~ should be available in each language, but `page_name.locale.erb` is unique to that language? 

Just a question, cause I was wondering the same when looking at a related PR

RonaldDijkstra:
See PR for details. Closed

PR #341, opened 2016-10-12
[READY] add unique for locale in frontmatter

RonaldDijkstra:
Checks if there's a unique_for_locale var in frontmatter. If that's true, it sends the flag link to the homepage of the other TLD. 

What do you think, is this sustainable enough @fatboypunk? 

Issue #239

### DIFF
--- a/source/localizable/hosting.de.erb
+++ b/source/localizable/hosting.de.erb
@@ -6,6 +6,7 @@ title:
 description:
   nl: Defacto biedt de mogelijkheid om CAPP LMS veilig en met betrouwbare security te hosten. Updates worden door Defacto geïnstalleerd en bijgewerkt.
   de: Defacto bietet die Möglichkeit, CAPP LMS mit zuverlässiger Sicherheit zu hosten. Updates werden von Defacto installiert und aktualisiert.
+unique_for_locale: true
 ---
 
 <section class="gray">
--- a/config.rb
+++ b/config.rb
@@ -273,7 +273,7 @@ helpers do
       img = image_tag("flags/#{lang}.gif", alt: flag_titles[lang])
       if is_blog_index?
         url = full_url("/blog/", lang)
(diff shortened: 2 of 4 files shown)


source/localizable/hosting.de.erb
```diff
@@ -6,6 +6,7 @@ title:
 description:
   nl: Defacto biedt de mogelijkheid om CAPP LMS veilig en met betrouwbare security te hosten. Updates worden door Defacto geïnstalleerd en bijgewerkt.
   de: Defacto bietet die Möglichkeit, CAPP LMS mit zuverlässiger Sicherheit zu hosten. Updates werden von Defacto installiert und aktualisiert.
```
dkhgh: I don't get this, if I remember correctly, the `name.locale.erb` files are already **exclusive** for that locale?
As opposed to the `name.erb` files which are translated through the i18n module.

RonaldDijkstra:
Correct, but the small flags link to 'name.locale.erb' even if the page doesn't have it's equivalent on the other TLD. Tried earlier to make it work based on sitemap search function but that couldnt be done. Earlier on @fatboypunk suggested a list, yesterday I got this idea for this issue.

source/localizable/hosting.de.erb
fatboypunk: yes they are exclusive, which means we don't want to link to a german page with the german flag if it isn't there.
so now we have this line, and i don't like it because it could be that every unique page should be added like this. https://github.com/DefactoSoftware/website/blob/4d1aa1db8fbcb18d432c541557439b8bd965323c/config.rb#L276

fatboypunk:
I like it, still have to check it though but looks good ^^

RonaldDijkstra:
Thanks!

RonaldDijkstra:
The biggest advantage is to prevent 404's

PR #358, opened 2016-11-01
[READY] Update text on buttons german website

RonaldDijkstra:
Team DE pointed out the Mehr [product_name] text on button isn't correct Deutsch, so I updated them to be correct. Needs to be reviewed by German team.


RonaldDijkstra:
It's reviewed by the German team and ready!

RonaldDijkstra:
Cool

PR #362, opened 2016-11-03
[READY] add blog about implementing a new system (in german)

RonaldDijkstra:
See title :)

Have to check if the link still works to the bigger version of the infographic after building.


fatboypunk:
adding 2 images?

PR #367, opened 2016-11-04
[READY] fix mobile menu

RonaldDijkstra:
- [x] check if var is not empty 
- [x] replace select to form js 
- [x] rename #id for select 

Oops 💩

### DIFF
--- a/source/javascripts/_forms.js.erb
+++ b/source/javascripts/_forms.js.erb
@@ -13,6 +13,14 @@
 $(function () {
   var $forms = $('form');
 
+  // Change option color on select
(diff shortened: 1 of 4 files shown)


source/javascripts/_forms.js.erb
```diff
@@ -10,9 +10,19 @@
 //   return attribute in document.createElement('input');
 // }
 
+
+
```
fatboypunk: don't need this :)

RonaldDijkstra: Thanks 👍

PR #369, opened 2016-11-07
[READY] Only init popup when present in DOM

### LINKED ISSUE #368: GA Send Event sends an event on all pages
Even when the page doesn't contain the pop-up (ie. blog indexes and posts), the event is send to Google Analytics on scroll. 

_popup.js
```
// Show popup after scroll
    this.$window.on('scroll', function () {
      if (!Defacto.popup.popupsShown && Defacto.popup.$window.scrollTop() > Defacto.popup.showScrollTop) {
        Defacto.popup.popupsShown = true;
        setTimeout(function () {
          $('.popup').addClass('popup-show');
          ga('send', 'event', 'popup', 'show', window.location.pathname);
        }, Defacto.popup.showDelay);
      }
    });
```

sn3p:
Fixes #368

@RonaldDijkstra this should fix it.

### DIFF
--- a/source/javascripts/_popup.js
+++ b/source/javascripts/_popup.js
@@ -36,8 +36,15 @@ Defacto.popup = {
   },
 
   init: function () {
+    var $popups = $('.popup');
+
+    // Check if a popup is present
+    if ($popups.length === 0) {
+      return false;
+    }
+
     // Remove popups the user has closed
-    $('.popup').each(function () {
+    $popups.each(function () {
       if (this.id && Defacto.popup.cookies[Defacto.popup.cookiePrefix + this.id] === Defacto.popup.cookieValue) {
         $(this).remove();
       }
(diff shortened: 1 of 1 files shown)


RonaldDijkstra:
Thanks

Tested and ready to go

source/javascripts/_popup.js
```diff
@@ -36,8 +36,15 @@ Defacto.popup = {
...
+    var $popups = $('.popup');
+
+    // Check if a popup is present
+    if ($popups.length === 0) {
+      return false;
+    }
+
     // Remove popups the user has closed
-    $('.popup').each(function () {
+    $popups.each(function () {
```
jessedijkstra: This isn't any different form what you were doing:
screen shot 2016-11-08 at 11 00 16

When popups is an empty array (the thing you're testing with the length check) the each doesn't do anything.

source/javascripts/_popup.js
```diff
@@ -47,6 +54,7 @@ Defacto.popup = {
     this.$window.on('scroll', function () {
```
jessedijkstra: Ah, the problem is this probably?

PR #376, opened 2016-11-15
[READY] add unique for locale to capp lms page

RonaldDijkstra:
defacto.nl/capp-lms doesn't exist on defactolearning.de,  
there it's /capp-bilden and /capp-entwickeln.  

-> Point it to the homepage


RonaldDijkstra:
Thx

PR #380, opened 2016-11-16
[READY] Fit hero better on mobile

RonaldDijkstra:
- [x] make the H1 & H2 better fit on mobile
- [x] make the hero buttons in the hero 100% wide on mobile (affects hero on LS page too)

### DIFF
--- a/source/stylesheets/components/_hero.scss
+++ b/source/stylesheets/components/_hero.scss
@@ -19,11 +19,12 @@
 
     button,
     .button {
-      margin: 0 0.5em 1.5em;
+      display: block;
+      margin: 0 0 1.5em;
 
-      @include media($medium-down) {
-        margin: 0 0 1.5em;
-        width: 100%;
+      @include media($medium-up) {
+        display: inline-block;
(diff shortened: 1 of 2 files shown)


source/stylesheets/components/_hero.scss
```diff
@@ -20,6 +20,11 @@
     button,
     .button {
       margin: 0 0.5em 1.5em;
+
+      @include media($medium-down) {
+        margin: 0 0 1.5em;
+        width: 100%;
```
sn3p: I'd use `display: block;` instead of `width: 100%;` in this case. This auto-fills the button so you won't need to reset its margin. Also lets consider a mobile-first approach where possible :)

```scss
button,
.button {
  display: block;
  margin: 0 0.5em 1.5em;

  @include media($medium-up) {
    display: inline-block;
  }
}
```

RonaldDijkstra: iphone5

Margin needs some tuning for smaller mobile screens. Other thing changed.

Issue #334, opened 2016-10-05
Use a "social proof" section on the Homepage


RonaldDijkstra:
As suggested before 
- clients 
- users
- courses created
- courses passed

Or different stuff. Just not just show logos.

dkhgh:
* Certificates passed

RonaldDijkstra:
Great

RonaldDijkstra:
![found](https://cloud.githubusercontent.com/assets/16101007/19806055/74d6e4d2-9d18-11e6-84cb-2e9361385f04.png)

Something like this. Found this on some website.

RonaldDijkstra:
Github Enterprise

![social-proof-github](https://cloud.githubusercontent.com/assets/16101007/20397082/b38cbb5e-ace8-11e6-9346-61cc445a45f2.png)

RonaldDijkstra:
100% geslaagde koppelingen!!!

RonaldDijkstra:
99.x % uptime?

RonaldDijkstra:
I love it when a plan comes together... 😆

lennartdefacto:
Is goed Hannibal Dijkstra :P

RonaldDijkstra:
Haha

PR #381, opened 2016-11-18
Add Proof section

RonaldDijkstra:
Issue #334 

- [ ] Concepting, how many items? Etc. 
- [ ] Dutch Content
- [ ] German Content 

### 2 concepts

proof

metrics-social-proof

### DIFF
--- a/source/localizable/_social-proof.erb
+++ b/source/localizable/_social-proof.erb
@@ -1,0 +1,5 @@
+<section class="social-proof">
+  <div class="row">
+    <%= partial "social-proof/social-proof" %>
+  </div>
+</section>
--- a/source/localizable/index.html.erb
+++ b/source/localizable/index.html.erb
@@ -31,4 +31,5 @@ sitemap_priority: 1
 
 <%= partial "events-overview" %>
 <%= partial "blog/featured" %>
+<%= partial "social-proof" %>
 <%= partial "clients/clients" %>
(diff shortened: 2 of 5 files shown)


AnneNH:
the second one

fatboypunk:
First one please it's more direct and gives a more direct sense of how much it is.
Maybe remove the `55 klanten`? doesn't seem as much like this not sure about the `klanttevredenheid`

RonaldDijkstra:
Thanks for your input

sn3p:
Nice, I love stats!

I prefer full numbers (first example). They are easier to read, but I'd try to evenly distribute (visually) the large numbers over the columns, e.g. in the middle or on the outsides.
The second example looks a bit more balanced, 3 columns is less cluttered.

I agree with fatboypunk that the amount of customers doesn't really add anything.
Lets figure out what data is interesting to communicate. Different data asks for a different presentation (form follows data).

What place you think we should show these stats on the website? Lets mock this into a screenshot :)

RonaldDijkstra:
Thank you!

RonaldDijkstra:
> What place you think we should show these stats on the website? Lets mock this into a screenshot :)

#### Just right above the logos of clients



#### Or after the solution boxes



I make sure I'll use real data soon.

GlennGeelen:
With the first image it feels more like you want to display defacto overall stats. 

With the second image I think it have to be relevant stats for the columns that are above them, else it might be confusing for the users.

So I think it depends on what stats you want to show, where to position the stats on the site.

sn3p:
Agree, what @GlennGeelen said :)

fatboypunk:
Jup, still like the first one though! 💃

RonaldDijkstra:
Maybe someday

PR #384, opened 2016-11-22
[READY] update ebook popup

### LINKED ISSUE #344: Pop-up still too big for some screens
Fix it with a media query based on height

RonaldDijkstra:
- [x] hide popup on small screens
- [x] fix scss linter issues (don't use id and color literals in scss)
- [x] ~~close popup on clicking anywhere else but the popup~~ (removed)
- [x] focus first (name) field of popup form on click

Fixes #344

### DIFF
--- a/source/javascripts/_popup.js
+++ b/source/javascripts/_popup.js
@@ -16,7 +16,7 @@ Defacto.popup = {
   // console.log(Cookies.get()); // show all coockies
   // Cookies.remove('defacto_popup-ebook', { path: '/' }); // remove ebook cookie
 
-  // Hide hightlight
+  // Hide highlight
   hide: function ($popup, cookieExpires) {
     if ($popup.length === 0) {
       return false;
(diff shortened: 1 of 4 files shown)


sn3p:
Would be nice to provide screenshots for visual changes. This way everyone can discuss without having to checkout.

source/javascripts/_popup.js
```diff
@@ -44,32 +45,45 @@ Defacto.popup = {
     }
 
     // Remove popups the user has closed
+    var popupCookie = Defacto.popup.cookies[Defacto.popup.cookiePrefix + this.id];
```
sn3p: `this.id` only works inside the each loop.

sn3p:
Looks good 👍  made some small changes, can you test and lemme know what you think.
Other that that, GTM!

RonaldDijkstra:
No we do not.

source/stylesheets/components/_popup.scss
```diff
@@ -37,6 +40,10 @@ $popup-mobile-mode: max-width 400px;
   width: 100%;
   z-index: 9999999999;
 
+  @include media($popup-hide-mode) {
+    display: none !important;
```
RonaldDijkstra: Need important to hide the popup, maybe remove it in the future, but not now.

RonaldDijkstra:
Click anywhere for closing was an extra method because on some smaller screens the close button was not visible. With the changes in this PR, the close button is always there or the popup is hidden. 

So.. Ready once again.

PR #385, opened 2016-11-23
Add Client Cases

RonaldDijkstra:
This is a (visual) concept for an 'extended' portfolio, as briefly mentioned in issue #331.

The content is all completely fake at this time, a video in the content is just an example how that would look and is for sure not a necessity. 

-  ~~Concepting (cases should not only be at the homepage, for example)~~
- [ ] Content Dutch cases
- ~~Content German cases (eventually)~~

## Homepage
homepage-cases

## Case 
case-concept


fatboypunk:
> Content German cases (eventually)

So we could do this in a seperate PR?

fatboypunk:
> Concepting (cases should not only be at the homepage, for example)

For now (this PR) this is good enough?

RonaldDijkstra:
For me it's a starting point to talk about how we could do this client cases. 

German is eventually for now for me because they have less time to create content.

fatboypunk:
> For me it's a starting point to talk about how we could do these client cases.

we could do that in the issue, I'd like to have this PR as small as possible because i like it and therefore we can push it sooner

RonaldDijkstra:
That's smart thinking.

RonaldDijkstra:
Maybe someday

PR #401, opened 2016-11-29
[WIP] change blogs author

RonaldDijkstra:
Do we want this to be changed?


GerwinV:
I don't think you can change the author of a blog. It was written by Joke, not someone else.

RonaldDijkstra:
Okay, in that case, shouldn't we have an 'archived' images folder?

fatboypunk:
> I don't think you can change the author of a blog.

Of course you can.
`source/blog/nl/2016-11-08-met-defacto-aan-de-top-in-de-ad-ziekenhuis-top-100.markdown` I think this should be changed anyway so the author can change with it.

For the other blog it might be nice to leave her as the creator of the blog

fatboypunk:
> Okay, in that case, shouldn't we have an 'archived' images folder?

We could we could also remove the missing icon image

RonaldDijkstra:
Nice idea

IwannaDJ:
I honestly think this should've never been a blog post. Maybe better to delete it or make a better story out of it? This is not blog worthy.

RonaldDijkstra:
This is not the case of this PR, should've been an issue in the first place, sorry. I feel for Marcel's solution with the removal of the replacement image and for removing the post. Tomorrow I'll make this right.

GerwinV:
Agree on that with @fatboypunk too.

PR #463, opened 2017-01-20
Update ruby

sn3p:
Update to Ruby version 2.3.3


fatboypunk:
why not 2.4?

sn3p:
> why not 2.4?

Because json gem (dependency) fails to load :(

PR #480, opened 2017-02-21
Comment out events partial

RonaldDijkstra:
Content is old, but want to preserve the partial to update the content quickly.


fatboypunk:
it's still in the git history right? no need to comment it out, just remove it :)

RonaldDijkstra:
True story true

PR #484, opened 2017-02-27
Capitalize the german blog tags

RonaldDijkstra:
According to @CathrinH it's better to capitalize the German tags, so that's what this is.


source/blog/tags/index.html.erb
```diff
@@ -14,7 +14,7 @@ pageable: false
     <% blog.tags.sort{ |a,b| a.to_s.downcase <=> b.to_s.downcase }.each do |tag, articles| %>
 
       <ol>
-        <h3>Tag: <%= link_to tag, tag_path(tag) %></h3>
+        <h3>Tag: <span class="tag"><%= link_to tag, tag_path(tag) %></span></h3>
```
fatboypunk: Why not do this in ruby instead of css?

RonaldDijkstra: First solution that came to mind. Will try ruby instead.

PR #492, opened 2017-03-08
Add https to the gravatar links

RonaldDijkstra:
External stuff should be https for when we make the switch to https.

When I just put this in the url variable

`https://secure.gravatar.com/avatar/#{hash}.png?d=404`

It doesn't return a 404 if the image/account doesn't exist. So this checks over http if the image/account exists and then makes a https URL


sn3p:
Ok for now, but can't we just kill Gravatar and get/add the avatars manually?
Gravatar slows down builds and breaks builds when the service is down (happened few times now).

RonaldDijkstra:
Yeah, will do that soon.

PR #503, opened 2017-03-22
Improve CAPP Agile Learning Page

RonaldDijkstra:
In this: 

- [x] Make sure German LS page redirects to something on the flag
- [x] A bit contrast on the 123
- [x] Make QR Clickable
- [x] Mobile order of features
- [x] Textual improvements
- [x] Make the nav-trigger white
- [x] Smaller hero quote on mobile
- [x] Improve up-to-date image

# Screenshots: 

### Desktop: 

capp-agile-groot

### Mobile: 

capp-agile-mobile


fatboypunk:
I like it, @sn3p anything to say about it or merge it?

source/localizable/capp-agile/_capp-agile.nl.erb
```diff
@@ -93,7 +94,7 @@
     </div>
 
     <div class="col col-img">
-      <%= image_tag "screenshots/capp-agile-qr-code.png", title: "On demand met QR-codes", alt: "On demand met QR-codes" %>
+      <a href="https://zorgdemo.learningspaces.io/s/qi4Dge-L_qY5aLzJnciT-A" target="_blank" rel="noopener"><%= image_tag "screenshots/capp-agile-qr-code.png", title: "On demand met QR-codes", alt: "On demand met QR-codes" %></a>
```
sn3p: This could use some newlines ;)

source/stylesheets/components/_dropdown-menu.scss
```diff
@@ -87,6 +87,21 @@
     }
   }
 
+  @at-root body.capp-agile-learning {
+    .nav-trigger {
```
sn3p: I'd try to avoid page specific styling in component like stylesheets, as it tends to get very complex. Instead we could overwrite this in `_capp_agile.scss`, or **even better** in this case, create a modifier class, e.g:

```css
.nav-trigger {
  ...

  &-white { ... }
}
```


So we can use `class="nav-trigger nav-trigger-white"` wherever we want 💃 

Also, I don't think we need `@at-root` here?

source/stylesheets/_capp_agile.scss
```diff
@@ -159,8 +188,18 @@ body.capp-agile-learning {
...
+        margin: 0 auto 1em;
+      }
+
       &-img {
         margin-bottom: 1em;
+        order: 1;
+      }
+
+      &-text {
+        order: 2;
```
sn3p: Awesome 👏

sn3p:
The numbered columns don't wrap properly for medium resolutions. This could be solved by using `width: 33%;` or go for a flex only solution.

screen shot 2017-03-23 at 20 06 10

source/localizable/capp-agile/_capp-agile.nl.erb
RonaldDijkstra: Done

source/stylesheets/components/_dropdown-menu.scss
sn3p: And something like this maybe?

config.rb
```rb
def nav_trigger_white?
  page.data.nav_trigger_white ? true : false
end
```

_header.erb
```erb
<a href="#" class="nav-trigger <%= "nav-trigger-white" if nav_trigger_white? %>">
  <span aria-hidden="true"></span>
</a>
```

RonaldDijkstra: D'accord! Made this

source/stylesheets/_capp_agile.scss
RonaldDijkstra: 👍

RonaldDijkstra:
Just 1 extra commit to make sure the numbered boxes don't get too small to read.

config.rb
```diff
@@ -352,6 +352,11 @@ def logo
     locale_link_to(img, "/")
   end
 
+  # Use frontmatter for white nav trigger on certain pages
+  def nav_trigger_white?
+    current_page.data.nav_trigger_white ? true : false
```
fatboypunk: what does `current_page.data.nav_trigger_white` return?

RonaldDijkstra: Thank you!

sn3p: `nil` or `true` in most cases, which will work.

fatboypunk: this is basically always not something you want
`? true : false` :)

PR #505, opened 2017-03-23
Add scroll effect to CTA button

RonaldDijkstra:
Adds a scroll effect and focus on the first field, after clicking the CTA. 

(CAPP Agile Learning Page)


source/javascripts/all.js
```diff
@@ -4,6 +4,8 @@
 //= require fitvids
 
 //= require _raf-polyfill
+//= require _scrollTo.min
```
sn3p: I think we should try to exclude external libs from the repo:

```
source "https://rails-assets.org" do
  gem "rails-assets-jquery.scrollTo"
end
```

source/javascripts/all.js
```diff
@@ -4,6 +4,8 @@
...
+//= require _scrollToForm
```
sn3p: And dasherize our partials ;)

source/localizable/capp-agile-learning.nl.erb
```diff
@@ -10,8 +10,7 @@ unique_for_locale: true
 
 <%= partial "capp-agile/capp-agile-testimonials" %>
 
-<section class="demo-request gray">
-  <a name="demo"></a>
+<section class="demo-request gray" id="form">
```
sn3p: Prefer a more specific name, e.g. `id="demo-request"`.

source/localizable/_hero_capp-agile.erb
```diff
@@ -6,7 +6,7 @@
 
       <h1><%= t("capp-agile.heading") %></h1>
 
-      <a class="button white-green" href="#demo" data-event="demo_request"><%= t("capp-agile.action") %></a>
+      <a class="button white-green" id="cta_button" href="#form" data-event="demo_request"><%= t("capp-agile.action") %></a>
```
sn3p: html/css is dasherized `id="cta-button"`

source/javascripts/_scrollToForm.js
```diff
@@ -0,0 +1,11 @@
+jQuery(function () {
+  $('#cta_button').click(function () {
+    $.scrollTo(this.hash, 600, {
+      onAfter: function () {
+        $('#name').focus();
+      },
+    });
+    $(this.hash).find('#form').text(this.title);
```
sn3p: Wait.. what is this sorcery?

1. This basically says `$('#form').find('#form')`, which will never be found.
2. And if this selector was found, the form content would be replaced with and empty string (the button doesn't have a title).

Also in my opinion `#name` ID is too generic. Also we could probably target the input by name:

```js
$(function () {
  'use strict';

  $('#cta_button').click(function () {
    var $target = $(this.hash);

    $.scrollTo($target, 600, {
      onAfter: function () {
        $target.find('input[name="name"]').focus();
      }
    });

    return false;
  });
});
```

source/javascripts/all.js
RonaldDijkstra: Dope

source/javascripts/all.js
RonaldDijkstra: Ok

source/localizable/capp-agile-learning.nl.erb
RonaldDijkstra: Done

source/localizable/_hero_capp-agile.erb
RonaldDijkstra: Ok :)

source/javascripts/_scrollToForm.js
RonaldDijkstra: Updated

RonaldDijkstra:
Change All The Things. 

Ready again

PR #506, opened 2017-03-23
Disable popup on homepage

RonaldDijkstra:
By request from Hans this disables the e-book pop-up on the homepage.


fatboypunk:
Do we have conversion numbers based on pages btw?

RonaldDijkstra:
I shared a few numbers.

PR #515, opened 2017-03-30
Turn hero and demo pic CAPP Agile Learning page

RonaldDijkstra:
Seems more natural to me. 

screenshot-hero


sn3p:
Yes way better!

One suggestion, can we photoshop the spaces index in the video? It still shows the old layout:

screen shot 2017-03-31 at 12 23 01

Issue #521, opened 2017-04-03
Locales: remove LS 


RonaldDijkstra:
Change LearningSpaces to CAPP Agile Learning.

See also PR #513

PR #523, opened 2017-04-04
Remove LS references from nl locale

RonaldDijkstra:
Half of #521 

Other half follows when we translate CAPP Agile Learning to German.


fatboypunk:
Why don't we leave the keys and change the values to _capp agile learning_?

sn3p:
> Why don't we leave the keys and change the values to capp agile learning?

I think all agile translations are already in place?

RonaldDijkstra:
This removes some stuff we do not use anymore.

PR #524, opened 2017-04-06
LunrJS Search

RonaldDijkstra:
LunrJS based search for Middleman. Works client side with a json file. 

Sources: 
https://github.com/manastech/middleman-search 
http://lunrjs.com/ 

## Todo for minimum viable:

- [x] Add lunr.js and middleman search gem 
- [x] Add locale page_title to search.json
- [x] Translations
- [x] Skip index pages from json (blog, tags, tag)
- [x] Search block in config.rb must have less lines :)
- [x] Localize stemmers & search languages
- [x] Styling
- [x] Search in main nav
- [x] Add active style to nav item when search page is active 
- [x] Search in mobile nav
- [x] Loading the json indicator
- [x] Make sure all indicated pages get properly stored to search.json (multiple blog issue) on localhost 
- [x] Fix version mismatch 0.7.0 -> 1.0.0 
- [x] Fix the hound issue with the multiple lines

- [ ] ~Disable input if search.json isnt downloaded yet~
When the download of search.json is finished the search executes!


source/localizable/_header.erb
```diff
@@ -31,6 +31,11 @@
         <li>
           <%= nav_link_to t("header.contact"), "/contact.html" %>
         </li>
+        <li>
+          <% link_to "/zoeken.html", class: "header-search" do %>
```
fatboypunk: Does this actually work? think it should be `<%=` right?

RonaldDijkstra: Works, the `<%=` actually doesn't in this case. 

Need to find a way to get an 'active' class in though

config.rb
```diff
@@ -126,6 +126,54 @@
...
+    url:     { index: false, store: true },
+    author:  { boost: 30 }
+  }
+
+  search.before_index = proc do |_to_index, to_store, resource|
+    if resource.data.title.is_a?(Hash) && resource.data.title[I18n.locale]
+      to_store[:title] = resource.data.title.send(I18n.locale)
+    end
+    if resource.data.search_skip?
+      throw(:skip)
```
sn3p: I think you should only use `throw` within a `catch` block.

source/blog/index.html.erb
```diff
@@ -5,6 +5,7 @@ description:
   nl: Ons blog staat vol met trends en verhalen over onze producten.
   de: Unser Blog ist voll von Trends und Geschichten über unsere Produkte.
 pageable: true
+search_skip: true
```
sn3p: Have you tried `index: false`? (see [this](https://github.com/manastech/middleman-search/blob/master/lib/middleman-search/search-index-resource.rb#L83))

config.rb
```diff
@@ -126,6 +126,54 @@
 
 activate :directory_indexes
 
+activate :search do |search|
+  search.resources = [
```
sn3p: Like we discussed, lets see if we can generate this list dynamically using `sitemap.resources`.

Also I'm curious, does this 'whitelist' works for translated urls?
Because `resource.path` returns "over-ons.html" for `:nl` and "de/uber-uns.html" for `:de`.

source/stylesheets/_header.scss
```diff
@@ -55,6 +55,10 @@
         &.active {
           color: $orange;
         }
+
+        &.header-search {
+          padding: 0 20px 0 0;
```
sn3p: You're only overwriting the left padding, so `padding-left: 0;` should be sufficient.

source/stylesheets/_search.scss
```diff
@@ -0,0 +1,44 @@
+@keyframes anim-rotate {
+  0% {
+    transform: rotate(0);
+  }
+
+  100% {
+    transform: rotate(360deg);
+	}
```
sn3p: indents 👮

sn3p:
Maybe something like this?

```rb
search.resources = []

ready do
  sitemap.resources.select { |r|
    r.path =~ /\.html/ &&
    r.metadata[:options][:lang] == I18n.locale &&
    !(r.path =~ /404.html/) &&
    !(r.path =~ /CNAME.html/) &&
    !(r.data["index"] == false)
  }.each do |page|
    pp page.path
    search.resources << page.path
  end
end
```

or 

```rb
search.resources = ["jobs/", "blog/"]

ready do
  sitemap.resources.select { |r|
    r.path =~ /\.html/ &&
    r.metadata[:options][:lang] == I18n.locale &&
    !(r.path =~ /CNAME.html/) &&
    !(r.path =~ /404.html/) &&
    !r.path.start_with?(*search.resources) &&
    !(r.data["index"] == false)
  }.each do |page|
    search.resources << page.path
  end

  pp search.resources
end
```

source/stylesheets/_search.scss
RonaldDijkstra: Weird.

config.rb
RonaldDijkstra: Suggestions?

source/blog/index.html.erb
RonaldDijkstra: Now I did and works, so the `:skip` is gone forever.

config.rb
RonaldDijkstra: NVM

RonaldDijkstra:
Nice going with the last suggestions 👍

config.rb
RonaldDijkstra: AFAIK Yes works

config.rb
```diff
@@ -126,6 +126,44 @@
 
 activate :directory_indexes
 
+activate :search do |search|
```
RonaldDijkstra: Cannot avoid this.

config.rb
```diff
@@ -126,6 +126,43 @@
...
+  search.resources = ["blog/", "jobs/"]
+
+  ready do
+    sitemap.resources.select { |r|
+      r.path =~ /\.html/ &&
+      r.metadata[:options][:lang] == I18n.locale &&
+      !r.path.start_with?(*search.resources) &&
+      !(r.data["index"] == false)
+    }.each do |page|
```
sn3p: We could do
```rb
pages = sitemap.resources.select { |r| 
  ...
}

pages.each do |page|
  ...
end
```

RonaldDijkstra:
When locally serving Middleman there's a problem with populating the search.json with pages with a blog (index, tags, jobs), if the json is downloaded through the javascript. When just typing the full url to search.json the file is just fine. 

This has something to do with the multiple blog we are running (blog/ and jobs/). With 1 of 2 blog types deactivated it's okay. 

The error in the terminal is (for the homepage in this example): 

`Error processing resource for index: index.html
You must either specify the blog name in calling this method or in your page frontmatter (using the 'blog' blog_name)`

When building there is no problem and the json is populated as expected. I just tested the search.json in the website from a build (and without search activated in the config) and it works with all the pages we need. 

Any suggestions how to make the localhost search.json as good as the one after building?

config.rb
RonaldDijkstra: Seems not to populate the search.json with any content.

sn3p: This works for me:

```rb
search.resources = ["blog/", "jobs/"]

ready do
  pages = sitemap.resources.select { |r|
    r.path =~ /\.html/ &&
    r.metadata[:options][:lang] == I18n.locale &&
    !r.path.start_with?(*search.resources) &&
    !(r.data["index"] == false)
  }

  pages.each do |page|
    search.resources << page.path
  end
end
```

config.rb
```diff
@@ -126,6 +126,43 @@
 
 activate :directory_indexes
 
+activate :search do |search|
+  search.resources = ["blog/", "jobs/"]
+
+  ready do
+    sitemap.resources.select { |r|
```
sn3p:
```rb
pages = sitemap.resources.select do |r|
  ...
end
```
🐶

sn3p:
> When locally serving Middleman there's a problem with populating the search.json with pages with a blog (index, tags, jobs), if the json is downloaded through the javascript. When just typing the full url to search.json the file is just fine.
> 
> This has something to do with the multiple blog we are running (blog/ and jobs/). With 1 of 2 blog types deactivated it's okay.
> 
> The error in the terminal is (for the homepage in this example):
> 
> Error processing resource for index: index.html You must either specify the blog name in calling this method or in your page frontmatter (using the 'blog' blog_name)
> 
> The pages do have a blog name specified in their frontmatter.
> 
> This is probably not a showstopper, because when building there is no problem and the json is populated as expected. I just tested the search.json in the website from a build (and without search activated in the config) and it works with all the pages we need.
> 
> Any suggestions how to make the localhost search.json as good as the one after building?

Two thoughts:

1. Create an issue at [middleman-search](https://github.com/manastech/middleman-search)
2. Could we exclude `blog/index.html` and `jobs/index.html` from the search? These pages are not at all interesting for search results.

RonaldDijkstra:
> Could we exclude blog/index.html and jobs/index.html from the search? These pages are not at all interesting for search results

Forgot to mention they already have index: false, except index.html.

source/jobs/index.html.erb
```diff
@@ -8,6 +8,7 @@ description:
   de: Wollen Sie bei einem der besten Unternehmen von Groningen zu arbeiten?
 pageable: true
 popup: false
+search_skip: true
```
sn3p: `index: false` right?

RonaldDijkstra: 🥇

sn3p:
> Forgot to mention they already have index: false, except index.html.

I found two things that trigger this error:

1. [blog/_featured.erb#L2](https://github.com/DefactoSoftware/website/blob/ronald/search/source/blog/_featured.erb#L2)
To fix this change `blog` to `blog("blog")`.

2. [blog/_blog-article-meta.erb#L13](https://github.com/DefactoSoftware/website/blob/ronald/search/source/blog/_blog-article-meta.erb#L13)
Triggered by `tag_path(tag)` so probably related to the tags.

sn3p:
> version mismatch: current 1.0.0 importing 0.7.0

The latest versions of [middleman-search](https://github.com/manastech/middleman-search/releases) uses Lunr 0.7.0, but the latest version of [Lunr](https://github.com/olivernn/lunr.js/releases) is at 2.0.1.
Maybe we could try updating it manually (e.g. using [rails-assets](https://rails-assets.org/#/components/lunr.js)).

sn3p:
Requested an update of lunr: https://github.com/manastech/middleman-search/issues/25

RonaldDijkstra:
Please review 👍

RonaldDijkstra:
**Future stuff:** 

- Search directly from navigation
- Make some metadata available when search (eg. specify what a result is, like a blog post or a page)
- Tweak LunrJS and hopefully update this to the latest version (with the gem)

source/localizable/_header.erb
```diff
@@ -158,6 +161,12 @@
             </div>
           </li>
 
+          <li id="search" class="dropdown search">
+            <% link_to "/zoeken.html", { class: "button" } do %>
+              <span class="icon-search"></span>&nbsp;&nbsp;<%= t("search.heading") %>
```
sn3p: Prefer layout through styling 🍺

fatboypunk:
🎉 🎉 🎉 🎉

RonaldDijkstra:
😆

PR #526, opened 2017-04-14
Add 70:20:10 to CAPP Agile Learning heading 

RonaldDijkstra:
Adds 70:20:10 to heading (as it's an important keyword which wasn't mentioned before in the page at all), makes the style more like the flyers and other stuff and a adds a slightly more subtle quote below the laptop image.


sn3p:
Screenshots would be nice 📷

sn3p:
Dope! Maybe we can make the slogan(s) bigger for large screens?

screen shot 2017-04-20 at 15 36 15

locales/nl.yml
```diff
@@ -77,7 +77,7 @@ nl:
     name: CAPP Agile Learning
     desc: "Performance support door 'kennis-on-demand'. Verbeter prestaties door medewerkers te voorzien van just-in-time kennis. Leg kennis simpel vast en deel handleidingen, protocollen, werkinstructies, QR-codes e.d."
     logo: logos/capp-agile-white.png
-    heading: "Performance support voor elke organisatie."
+    heading: "Performance support voor elke organisatie.<br>70:20:10 in de praktijk."
```
sn3p: I'm not a fan of combing slogans like this, but ok.

source/localizable/_hero_capp-agile.erb
```diff
@@ -4,7 +4,11 @@
 
       <%= image_tag "logos/capp-agile.svg", alt: "CAPP Agile logo", class: "logo" %>
 
-      <h1><%= t("capp-agile.heading") %></h1>
+      <h1>
+        <em>
+          <%= t("capp-agile.heading") %>
```
sn3p: We should avoid using [`em`](https://developer.mozilla.org/en/docs/Web/HTML/Element/em) for styling.

> Typically this element is displayed in italic type. However, it should not be used simply to apply italic styling; use the CSS styling for that purpose.

RonaldDijkstra:
How is your green so much more green than mine? :)

source/localizable/_hero_capp-agile.erb
RonaldDijkstra: D'accord

sn3p:
> How is your green so much more green than mine? :)

Weird, no difference on my screen :o

RonaldDijkstra:
Will find another way for this. Bye bye! 👍

PR #533, opened 2017-04-26
Replace CAPP Agile Learning hero image

RonaldDijkstra:
Make it (more) consistent with our print print materials


jessedijkstra:
👍

PR #534, opened 2017-04-26
Add width to reason boxes CAPP Agile page (support ie10+)

RonaldDijkstra:
The why choose CAPP Agile Learning boxes were not fit for IE10. Adding a % width fixes this.

## Without width: 
without

## With width:
with


fatboypunk:
@sn3p

PR #536, opened 2017-05-02
Add more options to conversion form CAPP Agile Learning

RonaldDijkstra:
Extends the form with a number of options, so it gives our visitors more options than just requesting a demo. This is probably good for conversions, because a demo could be a little too much to convert for some visitors. And it makes it more clear for our sales dept what people actually want with filling the form. 

## Screenshots: 
screenshot_more_options

## Checked checkbox: 
checked

This works in IE9+


fatboypunk:
What would this look like?

fatboypunk:
> Hou me op de hoogte via e-mail

Are we actually planning to send emails about it? otherwise it might be wise to have another default

RonaldDijkstra:
> Are we actually planning to send emails about it? otherwise it might be wise to have another default

I don't like checked by default checkboxes and removed it for now.

RonaldDijkstra:
@fatboypunk screenshot in the PR now.

fatboypunk:
@sn3p any thoughts on this?

RonaldDijkstra:
I'll propose this tomorrow in a meeting. Most likely some tweaks coming ;-)

fatboypunk:
> I'll propose this tomorrow in a meeting. Most likely some tweaks coming ;-)

Haha sure, i would like a default though maybe on more info. So that it doesn't add much of a barrier to the user.

RonaldDijkstra:
You're right, probably the first one and maybe make clear all options could be checked

RonaldDijkstra:
Removed the 'prijsopgave' option, now this is ready for review.

PR #541, opened 2017-05-09
Improve CMT page

RonaldDijkstra:
This page is getting old and can use a bit of ❤️ 

So: 

- [x] Improve copy
- [x] Update form
- [x] Update styling
- [x] Update images Agile
- [x] Maybe replace it to a better fitting place in the site, like themes or something


fatboypunk:
> Update images KPP, Agile

not doing this one?

RonaldDijkstra:
> not doing this one?

Just not the KPP one, because we need more time to set a decent environment. Will do this later. The LS ones are changed to Agile

RonaldDijkstra:
#558

PR #542, opened 2017-05-10
5 Column Footer (only NL first)

RonaldDijkstra:
This adds an extra column in the footer of the NL website, so we could cluster pages with a theme or topic which is not a product, service or an about us page. This feels more natural to me. For example pages about Convenant Medische Technologie, 70:20:10 and Performance Support. Tweaked some other things too. 

- [x] Add 5th column on dutch footer
- [x] Make the footer a bit more wide
- [x] Make link :hover `$orange` instead of white 
- [x] Replace all color literals with variables and fix other scss-linter stuff
- [x] Add search
- [x] Name the about us column `About Us` instead of `Defacto`
- [x] State the website is open source in the copyright frase and link to this repo 

At this moment we only have 1 theme page, but I am trying to release other themed landing pages really soon, because we want to be found at trending topics. 
That's why I put an extra example link in the screenshot below: 

## Screenshot: 

footer-impression


locales/nl.yml
```diff
@@ -123,10 +123,13 @@ nl:
     text_md: "Dit betekent niet dat je iets verkeerd hebt gedaan, het kan zijn dat we de pagina hernoemd of verplaatst hebben."
   footer:
     capp: "CAPP Leren & Ontwikkelen"
-    copyright_html: "© 1987 - %{year} Defacto"
+    copyright_html: "Deze website is %{link} © 1987 - %{year} Defacto"
```
sn3p: ❤️

source/stylesheets/_footer.scss
```diff
@@ -21,12 +35,18 @@
     }
 
     @include media($large-up) {
+      // Only NL has 5 columns yet
+      @at-root body.nl .footer .col {
+        @include span-columns(2.4);
```
sn3p: Doe we need `@at-root` here?
Doesn't this work?
```scss
@include media($large-up) {
  body.nl & {
    @include span-columns(2.4);
  }
}
```

RonaldDijkstra: It does.

locales/nl.yml
RonaldDijkstra: 💯

PR #546, opened 2017-05-11
70:20:10 themed landing page

RonaldDijkstra:
Dutch page about bringing the 70:20:10 principle in practice with our software solutions. Makes sense for us because advertising and a somewhat trending topic. And yes, this is my own copy :)

Todo
- [x] Styling
- [x] Copy


source/localizable/70-20-10.nl.erb
```diff
@@ -0,0 +1,44 @@
...
+popup: false
+---
+
+<div class="hero">
+  <div class="hero-inner">
+    <div class="hero-copy">
+
+      <h1>70:20:10 in de praktijk</h1>
+
+      <h2>Leren is veel meer dan het zenden van informatie</h2>
```
fatboypunk: If we want to add this page for the other languages in the future as well, we might already want to add everything to I18n

RonaldDijkstra: Done

PR #562, opened 2017-05-22
Add more contrast to dropdown

RonaldDijkstra:
On pages with a white background the dropdown menu lacks a bit of contrast. And a small update to spacing and the size of the arrow. 

Now

After PR
over defacto defacto


sn3p:
I like the border/dropshadow, but not sure about the arrow being opaque gray.
How about giving the arrow a border and shadow as well? A bit of a challenge in CSS but it can be done :)

RonaldDijkstra:
Yup, but can we do it like this first?

PR #566, opened 2017-05-23
Add background color to 702010 hero

RonaldDijkstra:
Noticed when the image isn't loaded yet, the background is (standard) $orange. No good for the button in the hero.


source/stylesheets/_70-20-10.scss
```diff
@@ -1,6 +1,8 @@
 body.x70-20-10 {
   .hero {
+    $background-color: #95c198;
```
fatboypunk: Do we want this here or in the variables?

RonaldDijkstra: Good point, done

PR #629, opened 2017-06-23
English: disable careers

RonaldDijkstra:
We don't have a translation of the jobs part at first so I want to disable this blog for the English locale. I can't seem to figure this out. I tried different things to not activate the jobs blog when the root_locale is :en. 

This part activates the blog: 

```ruby
activate :blog do |blog|
  blog.name = "jobs"
  blog.prefix = "jobs"
  blog.permalink = ":title"
  case root_locale
  when :nl
    blog.sources = "/nl/{title}.html"
  when :de
    blog.sources = "/de/{title}.html"
  end
  blog.paginate = false
end
```

I tried to make an if statement around the whole helper, or unless root_locale == :en on the activate line, but this gives errors and not a page not found on /jobs. 

Alternatively I could make copy on the page to state that 'we do not have any vacancies at this time'.


RonaldDijkstra:
It's fixed!

PR #639, opened 2017-06-26
English: contact

RonaldDijkstra:
Translates the contact page


locales/en.yml
```diff
@@ -82,6 +82,24 @@ en:
     desc: "Performance support through knowledge-on-demand. Enhance your employees’ achievements by providing them with just-in-time knowledge. Keep record of knowledge in a simple manner and share manuals, protocols or work instructions."
     logo: logos/capp-agile-white.png
   contact:
+    heading: "Hi! We’re so happy to meet you"
```
sn3p: `so`? Are we really THAT HAPPY? :D

**Alternatives:**
`Hi! We're happy to meet you`
`Hi, nice to meet you!`

locales/en.yml
```diff
@@ -82,6 +82,24 @@ en:
...
+    heading2: "We’re here to help, provide you with more information and answer your questions."
+    main:
+      heading: "How can we help?"
+      question:
+        heading: "More information"
+        info: "Do you have a question or would you love to know more about our software and services?"
```
sn3p: `like to know`?

locales/en.yml
sdegeus: Yeah ' nice to meet you' is better!

locales/en.yml
sdegeus: We'll turn it into like :-)

PR #664, opened 2017-06-28
English: organisation with an "s"

RonaldDijkstra:


source/localizable/_json_ld.erb
```diff
@@ -1,7 +1,7 @@
 <script type="application/ld+json">
   {
     "@context": "http://schema.org",
-    "@type": "Organization",
+    "@type": "organization",
```
sn3p: oops?

RonaldDijkstra: Yeah, it's with a Z for Schema.org, but forgot the capital O. 

http://schema.org/Organization

PR #670, opened 2017-07-03
English: update social links

RonaldDijkstra:
- [x] Add YouTube channel


RonaldDijkstra:
The social handles will ultimately change when they are ready, but for now this PR will do. It just adds the YT channel.

Issue #674, opened 2017-07-03
Homerun.co integration


RonaldDijkstra:
Only show vacancies for specific locale
Show summary (if possible)
Etc.

PR #675, opened 2017-07-03
Homerun integration (embed)

RonaldDijkstra:
No more blog for job postings, but using Homerun.co instead, to get a decent Recruitment Pipeline

We probably want to get access to the Homerun API to get more control over the way the vacancies are displayed. 

## Screenshot: 

werken bij defacto defacto

## Todo: 

- [ ] Improve copy of the page
- [ ] German translation of copy
- [ ] Display vacancies: we want title + summary, and **not** type and place

#674


fatboypunk:
So how much can we do with the texts, and how does end up in seo?

Also isn't it just possible with the api to forward formkeeps input to that pipeline?

RonaldDijkstra:
Must be possible. 
Would be great if we could get job posts on the website with the homerun api, then lead the applicants to the homerun forms. 

**Pros**
- advanced forms, like apply with linkedin and file uploads 
- recruitment pipeline with decent steps 
- more job post managing possibilities for HR

**Cons**
- losing some job pages on our website 
- embed as is is not sufficient, we want a title and summary, not a place and ft/pt 
- not optimal for the optimisation of the website, because of the way it's embedded right now

RonaldDijkstra:
This was way to old to be, gonna use this as a reference (see tag)

PR #704, opened 2017-07-07
English: client logos 

### LINKED ISSUE #619: English: which clients in client blocks?
We have two different client blocks.
One is used on several places, like the homepage.
The other one is on the CAPP Agile Learning page. 

I'd like to know which clients we want to show on these blocks. For now the same as on the Dutch website are in the blocks.

RonaldDijkstra:
Fix #619


source/localizable/capp-agile/_capp-agile-clients.en.erb
```diff
@@ -0,0 +1,33 @@
+<section class="gray clients">
+
+  <h2><%= t("capp-agile.clients_heading") %></h2>
+
+  <div class="logos">
+
+    <a href="https://www.pandadoc.com" target="_blank" rel="noopener">
+      <%= image_tag "clients/pandadoc.png", alt: "Pandadoc", title: "Pandadoc" %>
+    </a>
```
fatboypunk: but here we want the links?

PR #723, opened 2017-07-12
Test with HTML Proofer 

RonaldDijkstra:
Test after `build` whether all links are correct, pictures work, etc.

- [x] Add HTML proofer gem
- [x] Test + fix failures (#724)
- [x] Rewrite & configure test.rb 
- [x] Add rake tasks
- [x] Configure circle.yml

Source: https://github.com/gjtorikian/html-proofer


test.rb
```diff
@@ -0,0 +1,15 @@
+
```
fatboypunk: thix can be added :)

RonaldDijkstra:
For now I skip auto deploying. Ready for review.

PR #733, opened 2017-07-19
Flex the client logos

RonaldDijkstra:
- use flex box
- link
- colorize

example: 

logos


RonaldDijkstra:
`filter: grayscale();` doesn't work in IE11, but I don't think that's a real problem. The logos have an opacity setting anyway.

PR #734, opened 2017-07-20
Anonymize IP and add gaOptOut

RonaldDijkstra:
By request from Zee Germans here's an opt-out for the GA tracking cookie and we anonymize the IP from visitors. 

But only when `root_locale == 🇩🇪 `


source/localizable/_google_analytics.erb
```diff
@@ -1,10 +1,28 @@
...
+  <script>
+    var gaProperty = 'UA-6700447-1';
+    var disableStr = 'ga-disable-' + gaProperty;
+    if (document.cookie.indexOf(disableStr + '=true') > -1) {
+      window[disableStr] = true;
+    }
+
+    function gaOptout() {
+      document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/';
+      window[disableStr] = true;
```
sn3p: Just curious, are `gaOptout()` and `window[disableStr]` used by GA?

source/localizable/_google_analytics.erb
```diff
@@ -1,10 +1,28 @@
...
+  </script>
+<% end %>
+
 <script>
   (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
   (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
   m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
   })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
 
   ga('create', 'UA-6700447-1', 'auto', {'allowLinker': true});
```
fatboypunk: can't we use gaProperty here?

source/localizable/_google_analytics.erb
RonaldDijkstra: No, but.. 

https://developers.google.com/analytics/devguides/collection/analyticsjs/user-opt-out

PR #735, opened 2017-07-20
Add Datenschutz page

RonaldDijkstra:
With a text about that we use GA on the website, needed for Germany.

PR #737, opened 2017-07-20
Test: Don't check opengraph

RonaldDijkstra:
In this PR #735 there was a failing test of `og:url` because the page did not exist (yet), so let's not test the opengraph stuff?


RonaldDijkstra:
After deployment the page is there yes.

PR #747, opened 2017-07-25
Localize canonical in a helper

RonaldDijkstra:
So we can put canonical urls in based on the `root_locale`


config.rb
```diff
@@ -315,6 +315,19 @@ def meta_description(page = current_page)
...
   end
 
+  # Use frontmatter for canonical urls
+  def canonical
+    if current_page.data.canonical_url
+      html = ""
+      if current_page.data.canonical_url.is_a?(Hash) && current_page.data.canonical_url[I18n.locale]
+        url = current_page.data.canonical_url.send(I18n.locale)
+      elsif url = current_page.data.canonical_url
+      end
```
fatboypunk: this elsif can be removed right?

PR #754, opened 2017-07-26
change title media from #3 to #2

mildaskl:
Changed the title so the styles would be the same on every page


RonaldDijkstra:
Well done.

PR #761, opened 2017-07-31
Homepage (mobile) fixes

RonaldDijkstra:
- [x] Fit H1
- [x] Fit text in buttons
- [x] Update events partial (German) 
- [x] Add padding to headline div so it does not touch the edges


source/stylesheets/base/_typography.scss
```diff
@@ -21,7 +21,11 @@ h6 {
 
 h1,
 h2 {
-  font-size: 2.4em;
+  font-size: 2em;
```
sn3p: This applies to both hero and non-hero h1's right?
Is this the max we can go for mobile? feel a bit small, but maybe just need to get used to it :)

Some food for thought, I noticed many hero h1's have different sizes:

- Homepage = 2.8em (mobile = 1.6em)
- CAPP LMS = 2.4em (same for mobile)
- E-learning Starterkit = 4em (mobile = 1.6em)
- 360 Feedback = 4em (mobile = 1.6em)

Maybe we should make this a bit more consistent?

RonaldDijkstra: Yes we do

PR #763, opened 2017-08-01
Testimonials

RonaldDijkstra:
Side by side with `display: flex`, no matter how many testimonials there are.

- [x] Display flex on `featured-testimonials`
- [x] Replace colors and other vars to `variables.scss`


RonaldDijkstra:
Use `markitdown` instead of simple_format because `simple_format` was not parsing a link in the first german testimonial. 

Tested in IE10+, Edge, Chrome, Safari and FF.

PR #770, opened 2017-08-04
Alternate link helper

RonaldDijkstra:
If there's an alternate version of the page in another language add the link rel = alternate

`<link rel="alternate" hreflang="" href="" />`

Based on https://gist.github.com/xeraa/c30161bac25011a03058


config.rb
```diff
@@ -427,22 +427,21 @@ def country_flags
...
-        url = full_url locale_url_for(url, locale: lang)
+  # Returns alternate link tags for a given page
+  def alternate_link_tags(page)
+    link_tags = []
+    (langs - [I18n.locale]).each do |locale|
+      sitemap.resources.select do |resource|
+        if (resource.proxied_to == page.proxied_to &&
+            resource.metadata[:options][:lang] == locale)
+            href = resource.url.sub("/#{locale}/", full_url("/", locale))
+            link_tags << tag(:link, rel: :alternate, hreflang: locale, href: href)
```
fatboypunk: this line is indented once to much?

config.rb
```diff
@@ -427,22 +427,21 @@ def country_flags
...
+          next
```
fatboypunk: the `next` is to not add this to the output? can't we use reduce instead of select than?

sn3p: I think the `next` applies to the each here.
This means if the page is found for the iterated locale, it adds the link tag and moves to the next locale.

fatboypunk: It's inside the select right? so it's for the select
```ruby
[3, 2, 1, 2, 3, 2].each do |i|
  [i].select do |k|
    next if k == 1
    puts "shouldn't do this #{k}"
  end
  puts "does this #{i}"
end
```
shouldn't do this 3
does this 3
shouldn't do this 2
does this 2
does this 1
shouldn't do this 2
does this 2
shouldn't do this 3
does this 3
shouldn't do this 2
does this 2

fatboypunk: The indentation is worse now btw
```ruby
  def alternate_link_tags(page)
    link_tags = []
    (langs - [I18n.locale]).each do |locale|
      sitemap.resources.select do |resource|
        if (resource.proxied_to == page.proxied_to &&
            resource.metadata[:options][:lang] == locale)
          href = resource.url.sub("/#{locale}/", full_url("/", locale))
          link_tags << tag(:link, rel: :alternate, hreflang: locale, href: href)
          next
        end
      end
    end

    link_tags.join("\n    ")
  end
```
This makes it clear about what is in the statement of the if, and what's the code that gets executed

fatboypunk: @sn3p is this right or am i missing something?

RonaldDijkstra: Cool, thanks

RonaldDijkstra: Fixed it.

PR #777, opened 2017-08-10
Add translated German microlearning Blog

sdegeus:
Microlearning blog translated to German


RonaldDijkstra:
Did you check why the build fails?

PR #797, opened 2017-08-17
Set max nestingdepth to 6

RonaldDijkstra:
6 is suitable I think?


RonaldDijkstra:
Most of the times with a nestingdepth higher than 5 we merge it anyway..

PR #806, opened 2017-08-21
Fix errors

RonaldDijkstra:
- fixes a few misspelled images in blog
- adds a few missing images in blog
- adds https to the google doubleclick link


test.rb
```diff
@@ -11,5 +11,5 @@
   check_html: true, :validation => { :report_missing_names => false },
   check_favicon: false,
   check_opengraph: true,
-  alt_ignore: ["http://googleads.g.doubleclick.net/pagead/viewthroughconversion/1008438803/?value=0&guid=ON&script=0"],
+  alt_ignore: ["https://googleads.g.doubleclick.net/pagead/viewthroughconversion/1008438803/?value=0&guid=ON&script=0"],
```
RonaldDijkstra: But the URL is that long 🐶

fatboypunk: "my" \
"super" \
"long" \
"string"

PR #810, opened 2017-08-23
Client Cases

RonaldDijkstra:
Upgrading the references page to clients with use cases. 

- [x] Change references to clients
- [x] Add to main nav, update breakpoint
- [x] Redirect references to clients
- [ ] ...

klanten - defacto


RonaldDijkstra:
After a meeting today I'll use this as a reference for a different approach

PR #813, opened 2017-08-25
Use data ymls for references

RonaldDijkstra:
For easy maintenance.
Removed all the anchors. 

(We're going to redesign the clients part soon.)


RonaldDijkstra:
Thanks

PR #818, opened 2017-08-25
Add German knowledge transfer blog

sdegeus:
Still needs a last name and a translated introduction


sdegeus:
Date is correct and last name is added. It's ready to publish now.

PR #820, opened 2017-09-04
Remove test from deploy loop

RonaldDijkstra:
Optional, because test was sometimes blocking deploy without good reason, for example when we made a new page. 

Now a build test is optional and can be run with `rake test`. Runs a build before.


RonaldDijkstra:
PR's still get the full test. 👍

PR #830, opened 2017-09-07
Update homepage partials

RonaldDijkstra:
- [x] Replace includes in scss because we use autoprefixer
- [x] Add featured testimonial as a partial
- [x] Use one file for solutions instead of three (they are not that different anyway)
- [x] Put homepage partials in own folder for maintenance (the list is getting a bit long nowadays)


source/localizable/home/_solutions.erb
```diff
@@ -0,0 +1,54 @@
+<section class="solutions">
+  <div class="row">
+
+    <% unless root_locale == :de %>
```
sn3p: Could simplify this a bit I think:

```erb
<% if root_locale == :de %>
  ...
<% else %>
  ...
<% end %>
```

RonaldDijkstra: Oh ja :)

PR #832, opened 2017-09-07
Replace general partials

RonaldDijkstra:
For maintenance purposes the more general partials in a dedicated partials folder.
Some other ones replaced to their specific folders, like the hero of the CAPP Agile Learning page. 
Makes sense because the list with files was getting longer and longer.


RonaldDijkstra:
thx

PR #860, opened 2017-09-21
Save draft of new product blog

sdegeus:
Save as a draft for now. Just putting it here in case Hans wants this to go live in the next 3 weeks. 
Still needs images.


RonaldDijkstra:
Make this into a stale branch for future reference

PR #862, opened 2017-09-21
add elixierblog-de

DavidFritsch:
add elixirblog DE


DavidFritsch:
one time "Elixier" instead of "Elixir"

PR #866, opened 2017-09-25
Use a variable in the GA tracking code

RonaldDijkstra:
We lost verification of ownership in Google Search Console due to 'a malformed GA tracking code': 

> Verificatie mislukt voor https://www.defactolearning.de/ met de methode Google Analytics (6 minuten geleden). De trackingcode voor Google Analytics op uw site lijkt een onjuiste indeling te hebben.

Without using a variable for the GA id we can verify the ownership for search console again.


fatboypunk:
```js
 (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),

  ga('create', 'UA-6700447-1', 'auto', {'allowLinker': true});
  ga('require', 'linkid');
  ga('require', 'linker');
  ga('linker:autoLink', ['www.defacto.nl', 'jobs.defacto.nl', 'www.defactolearning.de', 'en.defacto.nl']);
    ga('send', 'pageview');
```
This is the output in the DOM, which is what google reads right? so it should be fine with a variable

RonaldDijkstra:
I temporary deployed this branch to live test if verification works without a variable (unfortunately there's no other way). 

When we put the variable back the verification breaks, because of a malformed GA snippet. So I think search console reads it as is, while the tracking works.

source/localizable/partials/_google_analytics.erb
```diff
@@ -1,11 +1,8 @@
-
-  <script>
-    var gaProperty = 'UA-6700447-1';
-
+<script>
   <% if root_locale == :de %>
 
     // Disable tracking if the opt-out cookie exists.
-    var disableStr = 'ga-disable-' + gaProperty;
```
sn3p: No clue why this fails, but apparently it does.

I think the best alternative is to create a config variable:

```rb
ga_code = "UA-6700447-1"
```

And use that in our template:

```erb
var disableStr = 'ga-disable-<%= config[:ga_code] %>';

ga('create', '<%= config[:ga_code] %>', 'auto', {'allowLinker': true});
```

RonaldDijkstra: Yeah!

PR #877, opened 2017-09-29
Update names of team members

RonaldDijkstra:
Just a minor thing that annoyed me :)

- [x] Remove extra space for names without a prefix
- [x] Add name as title to img
- [x] Helper 

Current output, notice the extra space before the lastname in the alt and h5

screen shot 2017-09-29 at 09 36 02

After this PR

screen shot 2017-09-29 at 09 37 15


source/localizable/over-ons/_team.erb
```diff
@@ -2,15 +2,22 @@
   <ol>
 
     <% data.team.sort_by { |p| p.lastname.downcase }.each do |person| %>
+
+      <% if person.prefix %>
+        <% full_last_name = person.prefix + " " + person.lastname %>
+      <% else %>
+        <% full_last_name = person.lastname %>
+      <% end %>
```
fatboypunk: I think this can become a function

RonaldDijkstra: Sure thing

PR #879, opened 2017-09-29
Update pics of team members

RonaldDijkstra:
- Using avatar property (simplify helper in the future)
- Probably conflicts with #878


RonaldDijkstra:
Resolved the conflict

fatboypunk:
> Add new team pics (except Lizette and new members)

So what do we do with the ones left behind?

RonaldDijkstra:
Update later on when Stéphanie is back, I think

PR #889, opened 2017-10-05
360 feedback copy update

mildaskl:


RonaldDijkstra:
Great job Milda, this feels way more as native english than it was.

PR #918, opened 2017-10-25
Add apply with homerun button

RonaldDijkstra:
Adding an option for an apply button which links to forms on Homerun, which have more options than FormKeep (upload a resume for ex.).

If there's an apply link in the frontmatter, apply button, else use the formkeep form

Other stuff: 

- make the formkeep form a partial 
- update the order a bit (sharing is not that important, actually nobody ever uses that)

First step towards transition completely to Homerun with integration in our website without javascript embed.


source/layouts/jobs_post_layout.erb
```diff
@@ -12,47 +12,32 @@
...
-        <strong><em><%= t("jobs.acquisition") %></em></strong>
-      </p>
     </div>
 
     <div class="col">
 
-      <%= partial "partials/social-share" %>
+      <% if current_article.data.apply_link %>
+
+        <% link_to "https://defacto.homerun.co/#{current_article.data.apply_link}/#{root_locale.to_s}/apply", class: "button apply", target: "_blank", rel: "noopener" do %>
```
fatboypunk: this is not printing?

RonaldDijkstra: eh i think it does..

RonaldDijkstra: screen shot 2017-10-25 at 17 11 09

sn3p: I would expect the same. Maybe make it `<%=` to be on the safe side.

source/layouts/jobs_post_layout.erb
```diff
@@ -12,47 +12,30 @@
...
-        <strong><em><%= t("jobs.acquisition") %></em></strong>
-      </p>
     </div>
 
     <div class="col">
 
-      <%= partial "partials/social-share" %>
+      <% if current_article.data.apply_link %>
+
+        <%= link_to t("jobs.form.apply"), "https://defacto.homerun.co/#{current_article.data.apply_link}/#{root_locale.to_s}/apply", class: "button apply", target: "_blank", rel: "noopener" %>
```
sn3p:
```rb
<%= link_to t("jobs.form.apply"),
      "https://defacto.homerun.co/#{current_article.data.apply_link}/#{root_locale.to_s}/apply",
      class: "button apply",
      target: "_blank",
      rel: "noopener" %>
```

For readability?

RonaldDijkstra:
Test fails on a blog from earlier on. And thanks for the remark @sn3p

PR #920, opened 2017-10-27
Escape le meta_description

RonaldDijkstra:
Someone thought it was a good idea to put quotes in the first lines of a blog post. 😄 

That breaks `meta_description`, so that needs to be escaped.


RonaldDijkstra:
Thanks Fastboypunk

PR #925, opened 2017-10-30
Deutsche Themen

RonaldDijkstra:
Adding Deutsche Themen pages. 

Need to find a better way to support only 2 out of 3 locales for a single page, because I think maintenance will get to hard with lots of different files for basically the same page (and the styling for example), but that's probably not in scope for this particular PR. 

- [x] Setup basic page structure
- [x] Add real content 70-20-10
- [x] Add real content Medizintechnik
- [x] Ignore alternate lang tag if page doesn't exist in other language


RonaldDijkstra:
Watch the test: 

`- ./build/de/70-20-10/index.html
  *  External link https://www.defactolearning.de/70-20-10/ failed: 404 No error
- ./build/de/medizintechnik-verbessern/index.html
  *  External link https://www.defactolearning.de/medizintechnik-verbessern/ failed: 404 No error
- ./build/de/themen/index.html
  *  External link https://www.defactolearning.de/themen/ failed: 404 No error`

These errors make sense, because it checks links that don't exist yet.

RonaldDijkstra:
> Need to find a better way to support only 2 out of 3 locales for a single page, because I think maintenance will get to hard with lots of different files for basically the same page (and the styling for example), but that's probably not in scope for this particular PR. 

Just ignore it in config.rb in the locale we don't need the page... lol

RonaldDijkstra:
@fatboypunk thanks!

RonaldDijkstra:
CircleCI:

```
- ./build/70-20-10/index.html
  *  External link https://www.defactolearning.de/70-20-10/ failed: 404 No error
- ./build/convenant-medische-technologie/index.html
  *  External link https://www.defactolearning.de/medizinproduktegesetz/ failed: 404 No error
- ./build/themas/index.html
  *  External link https://www.defactolearning.de/themen/ failed: 404 No error
```

fatboypunk:
> Ignore alternate lang tag if page doesn't exist in other language

This is fixed right?

RonaldDijkstra:
Yes we fixed that!

PR #934, opened 2017-11-06
Quality Monitor (NL)

RonaldDijkstra:
First NL only. 

- [x] Add basic page structure
- [x] Add to main nav
- [x] Add to footer
- [x] Add real logo
- [x] Add real hero image
- [x] Add real content 
- [x] Add real icon to main nav
- [x] Update conversion to productsheet download


sdegeus:
Content added - waiting for a text/spelling review, then it gets an update again.

locales/nl.yml
```diff
@@ -45,6 +45,7 @@ nl:
       capp: "Meer CAPP"
       cmt: "Meer CMT"
       quality: "Meer Kwaliteitspaspoort"
+      kwaliteitsmonitor: "Meer Kwaliteitsmonitor"
```
fatboypunk: why are the keys in dutch?

RonaldDijkstra: I know it's no good reason, but because it's the same for kwaliteitspaspoort. Should both be translated. But not now I think!

#983

RonaldDijkstra:
Ready for review I think.

locales/nl.yml
fatboypunk: But you started fresh with this one so you could've done it in one go

RonaldDijkstra: Could’ve but did not.

sdegeus:
Does someone have time to review this today?

source/localizable/kwaliteitsmonitor.html.erb
```diff
@@ -0,0 +1,43 @@
...
+description:
+  nl: Automatiseer het vinken
+lang_ignore: en, de
+---
+
+<div class="hero">
+  <div class="hero-inner">
+    <div class="hero-copy">
+
+      <%= image_tag t("kwaliteitsmonitor.logo"), alt: t("kwaliteitsmonitor.name"), class: "km-logo" %>
```
fatboypunk: we don't want a title on this?

sdegeus: Not that I know of. We can always add that later if we need it.

PR #939, opened 2017-11-07
Fix build errors due to wrong capp-lms links

RonaldDijkstra:
I removed a redirect from `/capp-lms` to `/capp-bilden` but forget there's some links to it and even an (old) redirect from the .NL website. 

When typing this I wonder if those redirects from `:nl/de/*` are still needed, I guess not...


RonaldDijkstra:
Yeah 🍏

PR #944, opened 2017-11-08
Use ignore_lang for flag and alternate links

RonaldDijkstra:
Specify more to ignore languages instead of just 1.

Can be used later on for the ignore build methode.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -465,7 +465,7 @@ helpers do
   end
 
   def lang_ignore?(page, locale)
-    page.data.lang_ignore && page.data.lang_ignore.include?(locale.to_s)
+    page.data.lang_ignore&.include?(locale.to_s)
   end
 
   def root_url?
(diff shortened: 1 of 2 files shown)


config.rb
```diff
@@ -462,6 +464,10 @@ def alternate_link_tags(page)
     link_tags.join("\n    ")
   end
 
+  def lang_ignore?(page, locale)
+    page&.data&.lang_ignore&.include?(locale.to_s)
```
fatboypunk: don't have to do page i guess, and are there pages without data?

RonaldDijkstra: Okay. 

And there are pages without page.data.lang_ignore yes

fatboypunk: yeah but also without page.data?

sn3p: `page.data` always exists so could remove the first `&`.

RonaldDijkstra: Ah cool thanks guys

RonaldDijkstra: We need 'page'

fatboypunk: it's not about the page it's about wether the safe operator is needed, you can use this:
`page.data.lang_ignore&.include?(locale.to_s)`

RonaldDijkstra: Aight!

PR #949, opened 2017-11-09
Merge hosting & security into 1 file

RonaldDijkstra:
Merge into 1 file for better maintenance.
Alphabetise the paths list in de.yml
Remove not longer needed redirects.


fatboypunk:
there's only stuff removed?

RonaldDijkstra:
only useless stuff, like 3 of the same frame pages for every locale. Maybe merge is a wrong word here? Or did I make a mistake? 

And ordered the paths in de.yml alphabetical because it annoyed me

fatboypunk:
> Maybe merge is a wrong word here?

yeah was confused by this

RonaldDijkstra:
I’ll update the commit message for better history then.

PR #951, opened 2017-11-13
Ignores through front matter

RonaldDijkstra:
No longer need to maintain a list of build ignores.


config.rb
```diff
@@ -33,43 +33,29 @@
...
+  ignores = ["/blog/de/*", "/blog/nl/*", "/jobs/*"]
+end
+
+ignores.each do |path|
+  ignore path
+end
+
+ready do
+  ignores = sitemap.resources.select do |i|
+    i.path =~ /\.html/ &&
```
fatboypunk: isn't i just the resource? why don't we name it so?

fatboypunk: Do we need to check if the path contains .html? or is just checking if there is a `lang_ignore` good enough?
What other pages do we state lang_ignore where we don't want to ignore it

config.rb
```diff
@@ -33,43 +33,29 @@
...
+      i.data&.lang_ignore.to_s.include?(I18n.locale.to_s)
+  end
+
+  ignores.each do |resource|
+    ignore resource.path
```
fatboypunk: Why do we loop through the resources again to ignore it? can't we do that in the loop above here?

config.rb
sn3p: If you skip the html check you'll also check agains images, scripts, stylesheets, fonts, etc.

A waste of resources I guess :D

fatboypunk: You're doing that here anyway right?

config.rb
RonaldDijkstra: No reason to double it no.

config.rb
fatboypunk: May leave this as is

config.rb
```diff
@@ -33,28 +33,24 @@
...
+    if resource.data&.lang_ignore.to_s.include?(I18n.locale.to_s)
+      ignores << resource.path
+    end
   end
 
-  ignores.each do |resource|
-    ignore resource.path
+  ignores.each do |path|
+    ignore path
   end
```
fatboypunk: I don't think this is needed
you could just do this where you're now doing `ignores << resource.path`

config.rb
```diff
@@ -48,8 +48,7 @@
...
 
   sitemap.resources.each do |resource|
-    if resource.path =~ /\.html/ &&
-       resource.data&.lang_ignore.to_s.include?(I18n.locale.to_s)
```
fatboypunk: no need to remove this

RonaldDijkstra: Haha ok just a sec :)

RonaldDijkstra: Removed the remove :)

PR #956, opened 2017-11-15
Use only one file for capp page

RonaldDijkstra:
Cleaning goes on. 

- Reduces files used to basically build the same pages
- Reduces styling output because all capp pages now use capp-lms as body class instead of unique classes for the same styling per page
- Localizes the heros (that's where /capp-bilden differs from /capp-lms the most)
- Links language flags from /capp-bilden to /capp-lms instead of homepage
- Uses rel=alternate from /capp-bilden to /capp-lms


source/stylesheets/_capp.scss
```diff
@@ -1,6 +1,4 @@
-body.capp-lms,
-body.capp-bilden,
-body.capp-entwickeln {
+body.capp-lms {
```
sn3p: Nice, but could also use this on those pages. This way we don't have to manipulate the page classes:

```scss
.hero-capp-lms {
   // ...
}
```

sn3p: I think we should try to keep `page_classes` clear of page specific rules as much as possible.

Alternatives:

- Use frontmatter to append classes. The only problem is that we currently don't have the resource available in the helper.
- Had an idea to use `@extend` but I think this has the same output as we had before:

```scss
body.capp-lms {
  // ...
}

body.capp-bilden {
  @extend .capp-lms;
}

body.capp-entwickeln {
  @extend .capp-lms;
}
```

RonaldDijkstra: I'd like this in an issue, do it later and ship this PR to production.

RonaldDijkstra: #957

sn3p: Roger.

PR #972, opened 2017-11-22
Don't show capp-lms option on German demo request page

RonaldDijkstra:
And shorten a bit.


source/localizable/demo-aanvragen.html.erb
```diff
@@ -33,10 +33,9 @@ popup: false
         <% end %>
         <select name="product" id="demo-select">
           <option selected><%= t("form.product") %></option>
-          <% if root_locale == :nl || :en %>
+          <% if root_locale == (:en || :nl) %>
```
fatboypunk:
```ruby
(irb(main):001:0> (:nl || :en)
=> :nl
```

PR #994, opened 2017-12-05
Add Productsheet to Qualitaetspass 

DavidFritsch:
- [x] Add Qualitaetspass to downloads.yml
- [x] Add content
- [x] Update form


data/downloads.yml
```diff
@@ -16,3 +16,5 @@ capp_productsheet:
   en: "/pdf/Productsheet_CAPP.pdf" # REPLACE_LATER!
 kwaliteitsmonitor_productsheet:
   nl: "/pdf/Productsheet_Kwaliteitsmonitor.pdf"
+kwaliteitspaspoort_productsheet:
```
fatboypunk: i think this is redundant and can the next line can be pased right under 
```yml
kwaliteitsmonitor_productsheet:
  nl: "/pdf/Productsheet_Kwaliteitsmonitor.pdf"
```

fatboypunk: my mistake

PR #995, opened 2017-12-05
Update listing of KM on product index

RonaldDijkstra:
Right colors, white logo.


locales/nl.yml
```diff
@@ -281,6 +281,7 @@ nl:
       text_md: "Automatiseer het vinken"
       image: screenshots/hero-kwaliteitsmonitor.png
     logo: logos/kwaliteitsmonitor.png
```
fatboypunk: this is still being used on other pages?

RonaldDijkstra: Yes!

PR #999, opened 2017-12-06
Rake post

RonaldDijkstra:
'Stolen' from Defacto devblog just for fun. 
https://github.com/DefactoSoftware/devblog

Still needs a check if the required params title and lang are specified or abort.


fatboypunk:
Do we now need a yml with the authors?

RonaldDijkstra:
No

PR #1010, opened 2017-12-12
Add newsletter subscribe box to blog posts

RonaldDijkstra:
This adds a **newsletter subscribe box** to blog posts. 

@sdegeus @DavidFritsch @SusanDefacto:

We probably do not want this box on every post, so it's optional with frontmatter: 

`subscribe_hide: true`


source/stylesheets/_blog.scss
```diff
@@ -91,6 +91,19 @@
       .col {
         margin: 0 auto;
         max-width: 960px;
+
+        &.post-subscribe {
```
sn3p: Think this can be defined after the `.col {}` ?
This prevents unnecessary nesting ;)

source/layouts/blog_post_layout.erb
```diff
@@ -11,11 +11,17 @@
   </header>
 
   <div class="row post-body">
+    
     <div class="col">
       <%= yield %>
     </div>
+
+    <%= partial "blog/blog-subscribe" unless current_article.data.newsletter_subscribe == false %>
```
sn3p: Little bit inceptions (but understand why). Maybe use `hide_subscribe` so we can do:

```erb
<%= partial "blog/blog-subscribe" unless current_article.data.hide_subscribe %>
```

RonaldDijkstra: Done

source/stylesheets/_blog.scss
RonaldDijkstra: Done

RonaldDijkstra:
@DavidFritsch @sdegeus to hide the newsletter subscribe box on a blog use `subscribe_hide: true` in frontmatter

PR #1016, opened 2017-12-18
Update ansprechpartnerin

RonaldDijkstra:


source/jobs/de/business-developer.html.markdown
```diff
@@ -32,4 +32,4 @@ Als Business Developer bist du verantwortlich für die strategische Weiterentwic
 * Eigenverantwortliche, herausfordernde Aufgaben
 * Luft, Liebe, Leben und Arbeiten in Groningen
 
-Haben wir Dein Interesse geweckt? Dann freuen wir uns über Deine aussagekräftige Bewerbung. Deine Ansprechpartnerin ist Esther Joosten, [e.joosten@defacto.nl](mailto:e.joosten@defacto.nl).
+Haben wir Dein Interesse geweckt? Dann freuen wir uns über Deine aussagekräftige Bewerbung. Deine Ansprechpartnerin ist Isabela Wojtowicz, [isabela@defacto.nl](mailto:isabela@defacto.nl).
```
DavidFritsch: Would it not be better to use the German defacto mail instead? isabela@defactolearning.de?

RonaldDijkstra: Yes.

PR #1023, opened 2018-01-02
Update jobs section

RonaldDijkstra:
- Remove unused stuff
- Remove redundant stuff (like vacancy in job titles, or summary on job preview)
- Give a more prominent position in the navs
- Simplify what can be simplified
- Add a text to the index about us as a arbeitgeber/werkgever


RonaldDijkstra:
Please review, it's already open for 7 days now.

PR #1060, opened 2018-01-31
Fix quote on cases

RonaldDijkstra:
- [x] Smaller line-height
- [x] Smaller position below quote

And remove unused class


RonaldDijkstra:
screen shot 2018-01-31 at 12 09 00

RonaldDijkstra:
Hound, pls wake up

source/stylesheets/_cases.scss
```diff
@@ -177,8 +177,9 @@ section.cases-overview {
   }
 
   .sidebar {
-    background: $white;
     text-align: center;
+    background: $white;
```
fatboypunk: why?

RonaldDijkstra: A deliberate mistake to see if the Hound would wake up.

RonaldDijkstra: It was hanging forever on 'waiting for status to be reported'

PR #1065, opened 2018-02-01
Remove media pages

RonaldDijkstra:
This page doesn't add any value for us and therefore should be removed.


source/localizable/header/_main-nav.nl.erb
```diff
@@ -130,10 +130,9 @@
           </h2>
 
           <ul class="links-list">
-            <li><%= nav_link_to t("header.jobs"), "/jobs/index.html" %></li>                                   
+            <li><%= nav_link_to t("header.jobs"), "/jobs/index.html" %></li>
```
RonaldDijkstra: Ah!

PR #1093, opened 2018-02-22
Cases page control 

RonaldDijkstra:
- [x] Add html and styling
- [x] Make it work
- [x] I18n


RonaldDijkstra:
screen shot 2018-02-22 at 12 49 37

Since we have only 2 cases, this always references to the same next and prev. But there will be more cases soon.

source/layouts/cases_post_layout.erb
```diff
@@ -100,4 +100,43 @@
...
 <%= partial "partials/social-share" %>
 
+<section class="page-controls">
+  <div class="row">
+
+    <div class="page-control-wrapper">
+
+      <%
+      prev_article = current_article.previous_article || blog.articles.first
+      next_article = current_article.next_article || blog.articles.last
```
fatboypunk: We could also check if they're the same to the current,

sn3p: Could skip prev article link if they are the same. Prolly need some stying adjustments.

RonaldDijkstra: Third case is in the making! Problem #solved 👍

source/layouts/cases_post_layout.erb
```diff
@@ -100,4 +100,43 @@
...
+              <%= t("cases.previous") %>
+          </span>
+          <h4 class="control-title">
+            <%= prev_article.title %>
+          </h4>
+        </a>
+      <% end %>
+
+      <% if next_article %>
+        <a href="<%= next_article.url %>" class="page-control page-control-next" rel="rel" title="<%= next_article.title %>">
```
sn3p: `rel="rel"`

RonaldDijkstra: relluh

PR #1095, opened 2018-02-22
Case Strukton and some other things

RonaldDijkstra:
This PR fixes some other thingies too, like a button on the cases-overview-context to all cases if the page is not /cases and i18n for the buttons

- [x] Basic content
- [x] Update styling
- [x] Update content


source/localizable/clients/_clients-cases-overview.erb
```diff
@@ -15,7 +15,15 @@ blog: cases
 
     <div class="case-wrapper">
 
-      <% page_articles.each do |article| %>
+      <%
+      articles = page_articles
+
+      if defined?(count) && count.is_a?(Integer)
```
fatboypunk: Why not just always add a count?

source/localizable/clients/_clients-cases-overview.erb
```diff
@@ -15,7 +15,15 @@ blog: cases
 
     <div class="case-wrapper">
 
-      <% page_articles.each do |article| %>
+      <%
+      articles = page_articles
```
fatboypunk: Not needed

source/localizable/clients/_clients-cases-overview.erb
RonaldDijkstra: Because on /cases i want it to show 'em all

fatboypunk: > Because on /cases i want it to show 'em all

So we can load an page_articles.size there?

RonaldDijkstra: I don't know.

RonaldDijkstra: Just used same thing as is used on the testimonials partials

RonaldDijkstra: This is NOT outdated, I didn't change this whatsoever.

source/localizable/clients/_clients-cases-overview.erb
RonaldDijkstra: Removed

source/localizable/clients/_clients-cases-overview.erb
RonaldDijkstra: I'll take a look though

sn3p: This was introduced to show only a fixed amount of testimonials (eg. for the homepage). The count was passed to the partial.
We probably don't have this application for cases, so could remove it I think.

source/localizable/clients/_clients-cases-overview.erb
sn3p: Removing this will break it? Think @fatboypunk means only assigning one variable to another doesn't make sense. But if you look at the bigger picture it does make sense to me:

```rb
articles = page_articles

if defined?(count) && count.is_a?(Integer)
  articles = articles.take(count)
end
```

But if we remove the count, we could just use `page_articles`.

source/localizable/clients/_clients-cases-overview.erb
RonaldDijkstra: I want a fixed amount of cases too on /klanten but not on /cases, so what to do with page_articles.size?

source/localizable/clients/_clients-cases-overview.erb
RonaldDijkstra: Reverted the removal.

source/localizable/clients/_clients-cases-overview.erb
sn3p: In that case this looks good to me 👍

RonaldDijkstra:
Ok and now wait for the real, approved content.

PR #1101, opened 2018-03-02
Update Case ZINN copy

RonaldDijkstra:
Remove and update some mistakies


RonaldDijkstra:
And at some links to products

PR #1108, opened 2018-03-06
Only use GA in production

RonaldDijkstra:
...


source/layouts/layout.erb
```diff
@@ -31,7 +31,7 @@
     <%= feed_tag :atom, "#{blog("blog").options.prefix.to_s}/feed.xml", title: "Atom Feed" %>
     <%= stylesheet_link_tag "main" %>
     <!--[if lt IE 9]><script src="//html5shiv.googlecode.com/svn/trunk/html5.js"></script><![endif]-->
-    <%= partial "partials/google_analytics" %>
```
fatboypunk: isn't circle ci running in test? maybe it's better to check if production?

PR #1133, opened 2018-03-28
Update clients

RonaldDijkstra:
- [x] Add SBA, timing, viecuri
- [x] Make a helper and reduce locale specific files, because 🇬🇧 is always same order as 🇳🇱


config.rb
```diff
@@ -605,6 +605,18 @@ def capitalize(tagname)
...
 
+  # Get clients
+  def clients
+    all_clients =
+      case root_locale
+      when :de
+        data["clients_de"]
+      else
+        data["clients"]
+      end
```
fatboypunk: This is just an `if ... else` right?

fatboypunk: This would also work
```rb
all_clients = :de ? data["clients_de"] : data["clients"]
```

RonaldDijkstra: Done, thx!

config.rb
```diff
@@ -605,6 +605,12 @@ def capitalize(tagname)
     tagname.gsub(/-[a-z]/, &:upcase)
   end
 
+  # Get clients
+  def clients
+    all_clients = root_locale == :de ? data["clients_de"] : data["clients"]
+    all_clients.clone
```
sn3p: Curious, why does this needs to be cloned? We're not manipulating the data arrays right?

RonaldDijkstra: Well was in the original code in the template so I thought that it's needed. Will have a look if that really is the case :wink:

PR #1156, opened 2018-04-12
Add new logo eko

### LINKED ISSUE #1153: Update logo EVK on clients page
See case right side: https://www.defactolearning.de/cases/evangelisches-krankenhaus-oldenburg/

That one is the one they use on their site. We use an old one on the clients page.

MarinkeDefacto:
Close #1153


RonaldDijkstra:
Merge when 🍏

PR #1190, opened 2018-06-12
Use right OG:URL for staging

RonaldDijkstra:
When Posting a link to HipChat (or other apps) it uses the open graph metadata. This makes sure staging.defacto.nl gets a proper link on said apps.


RonaldDijkstra:
screen shot 2018-06-12 at 12 28 42

In this example the image gets an anchor from the open graph metadata.

RonaldDijkstra:
Oh yes thanks brooo 👍

PR #1197, opened 2018-06-13
Remove events partial 🇩🇪

RonaldDijkstra:
No new content at this point, so better to remove it.


RonaldDijkstra:
`Sass::SyntaxError: File to import not found or unreadable: events.`

Ouch

PR #1229, opened 2018-07-11
Only import GA on production

### LINKED ISSUE #1118: GA if production
> isn't circle ci running in test? maybe it's better to check if production?

RonaldDijkstra:
Only import GA when build is for production. 

Close #1118


config.rb
```diff
@@ -13,10 +13,13 @@
 
 set :ga_code, "UA-6700447-1"
 
-# Set a target environment
+# Set target environments
 staging = ENV["STAGING"] == "true"
 set :staging, staging
 
+production = ENV["PRODUCTION"] == "true"
```
sn3p: Could name this var `is_production` or `production?`

PR #1247, opened 2018-07-23
Update some post styling

RonaldDijkstra:
- [x] Correct the space around social share section
- [x] Make all anchors in posts bold
- [x] Fit title on mobiles better because germans use ultra long words


RonaldDijkstra:
Thanks snep

PR #1240, opened 2018-07-17
Add blog conversion options + some blog styling updates

RonaldDijkstra:
- [x] Change subscribe to conversion
- [x] Fix styling
- [x] Add productsheet download option and make a folder
- [ ] Add a contact form option
- [ ] Add a request demo form option
- [ ] Add a link to CAPP LMS section option
- [ ] Select a conversion through frontmatter

...

PR #1249, opened 2018-07-23
Call-to-actions

RonaldDijkstra:
- [x] Remove blog newsletter subscribe
- [x] Make a folder with partials that can be used anywhere
- [x] Add a linear gradient to related posts background for more contrast
- [x] Move all call to actions to folder
- [x] Make new call to actions (capp zwolf for example)
- [x] Return of the newsletter subscribe
- [x] Stylings
- [x] Logics
- [x] Fallback
- [x] Rename some files
- [x] Squash


source/layouts/blog_post_layout.erb
```diff
@@ -21,14 +21,16 @@
...
-    <%= partial "blog/blog-subscribe" unless current_article.data.hide_subscribe %>
-
   </div>
 
 </article>
 
 <%= partial "partials/social-share" %>
 
+<% if current_article.data.call_to_action %>
+  <%= partial "call-to-actions/download-capp11-productsheet" if current_article.data.call_to_action == "capp11_productsheet" %>
```
sn3p: Nice, maybe could do something like this if we need more options:

```rb
def call_to_action(page_or_artice = current_page)
  case page_or_artice.data.call_to_action
    when "capp11_productsheet"
      "call-to-actions/download-capp11-productsheet"
    else
      "fallback?"
  end
end
```

```erb
<%= call_to_action("capp11_productsheet") %>
```

RonaldDijkstra:
If `current_article.data.call_to_action` is empty, then there's a fallback to `subscribe-newsletter`
If `current_article.data.call_to_action` is false, then no call-to-action will be shown
In all other cases `current_article.data.call_to_action` must match a partial in localizable/call-to-actions

RonaldDijkstra:
This was started in #1240 but going on with it I thought 'why not make all call-to-actions as partial available throughout the whole website?' and did just that.

config.rb
```diff
@@ -258,26 +258,26 @@ def to_store(resource)
...
-    html.remove_javascript_protocol = true
-    html.remove_http_protocol       = false
-    html.remove_https_protocol      = false
-    html.preserve_line_breaks       = false
-    html.simple_boolean_attributes  = true
-    html.preserve_patterns          = nil
-  end
-end
+# unless staging
+#   activate :minify_html do |html|
```
sn3p: Should these remain disabled?

RonaldDijkstra:
Oops

Groeten,

Ronald

Op 25 jul. 2018 om 07:43 heeft Matthijs Kuiper <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


@sn3p commented on this pull request.

________________________________

In config.rb<https://github.com/DefactoSoftware/website/pull/1249#discussion_r204986665>:

> -    html.simple_doctype             = false
-    html.remove_script_attributes   = true
-    html.remove_style_attributes    = true
-    html.remove_link_attributes     = true
-    html.remove_form_attributes     = false
-    html.remove_input_attributes    = true
-    html.remove_javascript_protocol = true
-    html.remove_http_protocol       = false
-    html.remove_https_protocol      = false
-    html.preserve_line_breaks       = false
-    html.simple_boolean_attributes  = true
-    html.preserve_patterns          = nil
-  end
-end
+# unless staging
+#   activate :minify_html do |html|


Should these remain disabled?

config.rb
RonaldDijkstra: Nope, did it for debuggin'

RonaldDijkstra: Fixed

RonaldDijkstra:
Hound.. 👎

Issue #1053, opened 2018-01-25
Cases + Clients


RonaldDijkstra:
Collect all issues thread for /cases

## Merged

- [x] Add own og:images instead of default #1277
- [x] Conversion forms #1263 
- [x] Display client logo in case header #1265 
- [x] link Producten & Diensten to the specific services/products #1261 
- [x] Position in sidebar smaller & gray #1060 
- [x] Smaller line-height for quote #1060 
- [x] add a divider/line between the body and sidebar content for mobile #1063
- [x] Share case on social media #1062 
- [x] /klanten and /cases have different spacing between title and paragraph #1061
- [x] Testimonials as part of clients #1077 
- [x] Cases on homepage #1145 
- [x] Next/prev or related  #1093
- [x] German Cases #1075

sn3p:
Maybe display the client logo in the case header as well?

RonaldDijkstra:
Sure, why not!

sn3p:
Not sure, but think the line-height for the quote is a bit to high?

sn3p:
Maybe add a divider/line between the body and sidebar content **for mobile**?

![screen shot 2018-01-25 at 16 35 45](https://user-images.githubusercontent.com/935529/35396742-d215592c-01ed-11e8-9b01-8b6b0cef92d5.png)

RonaldDijkstra:
Thanks for comments, will work on all stuff next week

DavidFritsch:
Maybe link Producten & Diensten to the specific services/products?

RonaldDijkstra:
Nice suggestion!

lennartdefacto:
Found this thread after also discovering that the 'Producten & Diensten' checklist items aren't linked to our products/services. Is this something to pick up now or implement it simultaneously with the redesign of the website?

RonaldDijkstra:
Yes

RonaldDijkstra:
All done... waiting for reviews ;-)

RonaldDijkstra:
Add own og:images instead of default

RonaldDijkstra:
All is good.

PR #1263, opened 2018-08-01
Cases Call-to-Actions

RonaldDijkstra:
- [x] Make call-to-actions a partial 
- [x] Make case call-to-action (appointment-case) 

#1053


RonaldDijkstra:
Too early approved

PR #1308, opened 2018-08-28
Check for notification so javascript works

RonaldDijkstra:
No js at english website right now because we didn't consider this. But now we do.


source/javascripts/_privacy-notification.js
```diff
@@ -4,6 +4,10 @@ $(document).ready(function() {
   var cookieName = "privacy_notification_dismissed";
   var cookieValue = Cookies.get(cookieName);
 
+  if (!notification) {
+    return false;
+  }
+
   console.log(typeof cookieValue);
```
sn3p: Can remove this 😄 debug leftover

RonaldDijkstra: But I LIKE that.

PR #1318, opened 2018-09-03
Update Workshop form endpoint

RonaldDijkstra:
Custom endpoint to be able to send a custom mail instead of the standard contact one.


RonaldDijkstra:
screen shot 2018-09-03 at 14 18 45

PR #1391, opened 2018-10-08
Update page_titles

RonaldDijkstra:
Add 'Software' and use snake_case


RonaldDijkstra:
"Developing People" was and is beautiful, but doesn't say much about our products and offerings.

Issue #1388, opened 2018-10-08
No cookies solution for Vimeo and YouTube embeds


RonaldDijkstra:
Cause nobody likes cross site tracking.

PR #1405, opened 2018-10-09
Use no-cookie url for youtube embed

RonaldDijkstra:
This doesn't stop all, but sends a lot less cookies. 

Part of #1388


RonaldDijkstra:
Read more https://www.ghacks.net/2018/05/23/why-you-should-always-use-youtubes-privacy-enhanced-mode/

PR #1410, opened 2018-10-09
Update posts typography and width

RonaldDijkstra:
For readability


source/stylesheets/_jobs.scss
```diff
@@ -1,18 +1,16 @@
-body.jobs {
+body.jobs_index {
+
   .col {
     margin: 0 auto;
-    max-width: 960px;
+    max-width: 768px;
```
sn3p: We could make a variable for this, just a thought:

```css
$post-max-width: 768px;
```

RonaldDijkstra: Just a good thought

PR #1417, opened 2018-10-10
Add schema.org markup to Gebruikersbijeenkomst

RonaldDijkstra:
For maintenance purposes I just put this in the main body of the page, so when an event gets removed the schema won't be forgotten.

https://www.schema.org


RonaldDijkstra:
Part of #1415

PR #1449, opened 2018-10-16
Use a less cookie option for vimeo

RonaldDijkstra:
`allowscriptaccess="never"`

Should make sure there are no cookies from Vimeo at all, but I'm not convinced.


RonaldDijkstra:
#1388

PR #1474, opened 2018-10-18
Fix links to kpp in cases content

RonaldDijkstra:
Use a / at the end of internal linkies


RonaldDijkstra:
screen shot 2018-10-18 at 16 36 48

PR #1478, opened 2018-10-18
Add a page about Service & Support

### LINKED ISSUE #496: Add a page about Service & Support
Graag een link naar het Defacto Portaal op de website plaatsen. Je mag zelf bedenken waar :)
Wanneer je geen idee hebt: Vanuit 'contact' in het rijtje? zie printscreen.

Alvast bedankt!
contact defacto

RonaldDijkstra:
A longtime wish so let's do this. It makes a bit more clear what a (potential) client can expect from our Service & Support Team. 

- [x] Register a page
- [x] Register a stylesheet
- [x] Add contents 
- [x] Add to navigations

Close #496


source/localizable/support.html.erb
```diff
@@ -26,11 +26,45 @@ lang_ignore: de, en
...
+      <p>
+        Uiteraard kunt u ook via e-mail en telefonisch contact met ons opnemen. Bel tijdens kantooruren 050 - 3144 832 of mail naar <a href="mailto: info@defacto.nl">info@defacto.nl</a>. U kunt ons ook mailen via <a href="/contact/#contact-form">het contactformulier.</a>
+      </p>
+
+      <h3>
+        Feature requests
+      </h3>
+
+      <p>
+        Ons service & support team staat voor u klaar voor de dagelijkse ondersteuning bij het gebruik van onze software. Feature requests verlopen via onze consultants.
```
RonaldDijkstra: @lennartdefacto we need better copy here.

source/localizable/support.html.erb
```diff
@@ -0,0 +1,114 @@
...
+      <p>
+        Via het <a href="https://defactoportaal.cappagile.com" target="_blank">Defacto Portaal</a>, waarvoor we handig gebruikmaken van de mogelijkheden van CAPP Agile Learning, publiceren we alle nodige en nuttige documentatie over onze software, zoals release notes, handleidingen en nog veel meer.
+      </p>
+
+      <h3>
+        E-mail of telefoon
+      </h3>
+
+      <p>
+        Uiteraard kunt u ook via e-mail en telefonisch contact met ons opnemen. Bel tijdens kantooruren 050 - 3144 832 of mail naar <a href="mailto: info@defacto.nl">info@defacto.nl</a>. U kunt ons ook mailen via <a href="/contact/#contact-form">het contactformulier.</a>
```
sn3p:
```
mailto:info@defacto.nl
```

RonaldDijkstra:
Service & Support checked contents and approved the content

source/stylesheets/pages/_support.scss
```diff
@@ -0,0 +1,68 @@
...
+
+      img {
+        @include avatar-size(120px);
+        margin-bottom: 0.2em;
+      }
+
+      h5 {
+        font-weight: 700;
+
+        span {
```
sn3p: Why not use a `div` in the template? Then you don't need styling :)

source/stylesheets/pages/_support.scss
```diff
@@ -0,0 +1,68 @@
...
+
+  section.support-team {
+    background: $light-gray-two;
+
+    h2 {
+      text-align: center;
+    }
+
+    .row {
+      max-width: 640px;
```
sn3p: Maybe we should define variables like `$row-wide`, `$row-default`, `$row-narrow`?
Or add similar helper classes.

This way we have a system instead of random widths all over.

source/stylesheets/pages/_support.scss
RonaldDijkstra: Because reasons. ;-) Or: i dunno

source/stylesheets/pages/_support.scss
RonaldDijkstra: #1522

source/stylesheets/pages/_support.scss
RonaldDijkstra: #1523

RonaldDijkstra: Because it isn't semantically correct is the answer here. 

screenshot 2018-11-19 at 14 35 43

PR #1532, opened 2018-11-01
Don't set image in asset_hash ignores

RonaldDijkstra:
Image doesn't exist, so it doesn't need to be ignored here


sn3p:
The image should nog get hashed, it's used in https://codepen.io/snap/full/NGxgLr

PR #1535, opened 2018-11-01
Restructure and refactor Helpers 

RonaldDijkstra:
All work done in #1434 in smaller bits for better reviewing.

This is reworking the Helpers to move them from Config.rb to a designated folder and on the side refactor them a bit. For me this is pretty huge because I learn a lot from doing this.  👍 

- [x] Add language switcher #1537
- [x] Add content helpers #1541 
- [x] Add meta helpers #1546  
- [x] Add navigation helpers #1543 
- [x] Add page class helpers #1547 
- [x] Add post helpers #1548 
- [x] Add misc helpers #1549 
- [x] The other things #1549

PR #1537, opened 2018-11-01
Add language switcher

RonaldDijkstra:
Adds the `language_switcher` helper and all helpers it depends on. 

Based on #1535


helpers/miscellaneous.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Helpers for miscellaneous things like styling
+module Miscellaneous
```
sn3p: `module Locale` maybe?

helpers/current_page.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+# Helpers for current page functions
+module CurrentPage
```
sn3p: `module Page` ?

Issue #1355, opened 2018-10-03
Update main navigation


RonaldDijkstra:
Make it...

- more smooth 
- functionally better
- work better on mobile

And add new pages if any.

RonaldDijkstra:
Preview: 

https://staging.defacto.nl

RonaldDijkstra:
"Striped"

RonaldDijkstra:
Preview is gone atm.

PR #1558, opened 2018-11-07
Fix svg inlining (logos + icons) 

### LINKED ISSUE #1550: Clean SVG (logos)
They can be compressed a bit more by removing useless data: 

`<!-- Generator: Adobe Illustrator 21.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->`

RonaldDijkstra:
- [x] Strip the logos (no more useless metadata) close #1550 
- [x] Strip the icons and other svgs
- [x] Add `svg_inline` helper (from #1355)
- [x] Inline all svg images


source/localizable/contact/_contact.md.erb
```diff
@@ -7,7 +7,7 @@
 
     <div class="col">
       <div class="contact-image">
-        <%= image_tag "icons/icon-contact.svg", title: t("contact.main.question.heading"), alt: t("contact.main.question.heading") %>
+        <%= inline_svg "icons/icon-contact.svg", title: t("contact.main.question.heading") %>
```
sn3p: What is the reason for inlining these?

RonaldDijkstra:
Reduce requests

sn3p:
> Reduce requests

Just once, and after that it's cached right?
I'm wondering if reducing requests is an advantage in this case.

Issue #1383, opened 2018-10-05
Test and fix accessibility


RonaldDijkstra:
Where applicable

RonaldDijkstra:
Return of the outline.

RonaldDijkstra:
https://a11yproject.com/

RonaldDijkstra:
https://a11yproject.com/checklist

RonaldDijkstra:
#1438

PR #1562, opened 2018-11-09
Use what-input to style :focus

RonaldDijkstra:
Dont like what happens when there's a focus style and a click by a mouse on a link. So introducing [what-input](https://github.com/ten1seven/what-input) to explicitly target keyboard input


RonaldDijkstra:
screenshot 2018-11-09 at 14 29 36

Big fat outline to make it very clear where the user is

RonaldDijkstra:
#1383

Issue #1381, opened 2018-10-04
Check and fix html document structure(s)


RonaldDijkstra:

RonaldDijkstra:
H1's on cases don't contain anything according to the spiders

RonaldDijkstra:
We didnt make many mistakes in all that time 👍

RonaldDijkstra:
Should we use `<html lang=en>`if page is written in English, if the tld is another than en.defacto.nl?

RonaldDijkstra:
https://www.w3.org/International/questions/qa-html-language-declarations

PR #1584, opened 2018-11-15
Add heading to clients section

RonaldDijkstra:
And a button to `/clients`

Part of #1381


sn3p:
Screenshots would be nice, I'm lazy and don't want to check out every PR 😄

RonaldDijkstra:
I am actually researching a way to make deploy reviews available, so don't you worry about that ;-)

In the meantime, here's a screenshot

screenshot 2018-11-22 at 10 50 05

PR #1592, opened 2018-11-16
Remove past Defacto event from eventpages

lennartdefacto:
- [x] Remove event from NL page
- [x] Remove event from DE page


lennartdefacto:
Ronald has approved this while standing over my shoulder

PR #1599, opened 2018-11-20
Wrap content in actual language

RonaldDijkstra:
Some pages, like [this one](https://www.defacto.nl/open-source/) actually have a different language for the content than is defined in the `html` tag. This wraps a `<span>` with a `lang=` tag around said content or in the article tag on posts. 

Part of #1381


RonaldDijkstra:
Failing build is a 302 from Facebook

helpers/locale.rb
```diff
@@ -11,4 +11,11 @@ def other_locales
   def lang_ignore?(page, locale)
     page.data.lang_ignore&.include?(locale.to_s)
   end
+
+  # Use a lang tag if the content has a different language
+  def content_language(page = current_page)
+    "lang = #{page.data.content_lang}" if
+      page.data.content_lang &&
+      page.data.content_lang != I18n.locale.to_s
```
fatboypunk: Don't think we need this, if the content_lang is given it should always be the content_lang right?

RonaldDijkstra: Although I guess it doesn't hurt, stating twice what the language is is redundant and is why I check against the current locale. Probably not the nicest way.

For example the `<html lang="en">` tag on https://en.defacto.nl/open-source/ would be stated again in the `<span>`. 

Maybe I should add real languages per locale in frontmatter, but is redundant again:

```ruby
content_lang:
  de: en
  en: 
  nl: en
```

RonaldDijkstra:
I try to deal here with something that should not be the case in the first place. No single page should contain more than one language for big parts of the content.

Open source page should be translated (we can debate about the repo descriptions). 
Dev blogs should be moved to dev blog.

helpers/locale.rb
sn3p: > ... is is redundant

🤣

PR #1605, opened 2018-11-22
Update product index(es)

RonaldDijkstra:
Use svg where possible
Update some filenames


source/stylesheets/pages/_producten.scss
```diff
@@ -101,6 +101,10 @@
 .capp-lms {
   .left {
     background: $pink;
+
+    img {
+      width: 210px;
+    }
```
EasterPeanut: what image is this?

Issue #1384, opened 2018-10-05
Restructure stylesheets and config


RonaldDijkstra:
- [x] All helpers to a Helpers folder with a bit of refactoring #1535 
- [x] Stylesheets are getting out of control, restructure that #1423 
- [x] Classes
- [x] Settings

PR #1619, opened 2018-11-26
Update config again

RonaldDijkstra:
- [x] Add some comments
- [x] Move TextExceptionLocalizationHandler to /lib
- [x] Redirect with data files

Part of #1384, which I think after this PR should be closed.


config.rb
```diff
@@ -95,52 +101,23 @@
...
+    data.redirects_de
+  when :en
+    data.redirects_en
+  end
 
-# Set the timezone
-Time.zone = "CET"
+# Redirect each defined old_path to new_path
+data_redirects.each do |redirect|
+  redirect redirect.old_path.to_s, to: redirect.new_path.to_s
```
sn3p: 💯👍

data/redirects_de.yml
```diff
@@ -0,0 +1,16 @@
+- old_path: "events.html"
```
sn3p: Alternative (shorter) format could be:

```yaml
- from: "events.html"
  to: "veranstaltungskalender.html"
```

RonaldDijkstra: Had this earlier but thought it wasn't descriptive enough ;-)

PR #1623, opened 2018-11-28
Remove former employee from About Us page

lennartdefacto:
- [x] Remove former employee from About us page
- [x] Remove employee image


lennartdefacto:
Left the image in because of the following issue: #1624

RonaldDijkstra:
Lets wait till 1-12

PR #1644, opened 2018-12-05
Fix post header padding

RonaldDijkstra:
By removing the bigger padding that doesn't have the $header-height


RonaldDijkstra:
Before
screenshot 2018-12-05 at 15 43 10

After
screenshot 2018-12-05 at 15 43 23

PR #333, opened 2016-10-05
[READY] Homepage hero behind menu bar

RonaldDijkstra:
A hero image on the homepage which is behind the menu too 

- [x] Behind menu on homepage, with an inverted logo on the homepage
- [x] German hero image which is more 'health care'

PR #1706, opened 2019-02-21
Add migration page (or: how to make a landing page with Middleman)

### LINKED ISSUE #1705: Add 'migratiekit' landing page
We need a landing page for newsletter readers to view additional information about the migration and download the product sheet (CTA with form and PDF download).

RonaldDijkstra:
Close #1705 

- [x] Add page
- [x] Add structure
- [x] Add stylesheet
- [x] Add download
- [x] Add content
- [x] Optimize download
- [x] Get feedback on and update content 
- [x] Decide whether we want this page indexed in Google or not and set the frontmatter meta accordingly


RonaldDijkstra:
With `lang_ignore: en, de` in the frontmatter we make sure the page won't be build in that languages, as we don't need those. 

With `index: false` we make sure the page won't get added to `sitemap.xml` (default: true)

With `robots: noindex, nofollow` we make sure the page doesn't get indexed (default: `noydir,noodp,index,follow`. 

Defaults mean when you remove the lines from frontmatter, default will be used.

RonaldDijkstra:
In the stylesheets we use variables, which can be found in `/source/stylesheets/config/_variables.scss`. So in this case we use `color: $dark-gray;`, which resolves to `$dark-gray: #333;`

RonaldDijkstra:
As I do not have a (PDF) file of the sheet right now I added a basic form for contact request about migration to the latest greatest. You have to figure out how you want the 'conversion' to be. Do we want to ask existing customers for their data again? Or do we just want to link to a PDF file without a form first?

Content is in a markdown file: `source/localizable/migration/_content.nl.md.erb` and gets parsed from back to front. So first ERB, then Markdown and then for NL only. WIth the underscore in the filename we make sure the partial doesn't get build as a page. When we do not use an underscore, the page `migration/content/index.html` would be build.

data/downloads.yml
```diff
@@ -17,3 +17,5 @@ capp_productsheet:
 kwaliteitsmonitor_productsheet:
   nl: "/pdf/Productsheet_Kwaliteitsmonitor.pdf"
   de: "/pdf/Productsheet_Qualitaetspass_Qualitaetsmonitor.pdf"
+migratiekit_productsheet:
+  nl: "/pdf/Productsheet_Migratie.pdf"
```
RonaldDijkstra: To be honost, I do not like underscores in filenames. They don't mean anything and bots will ignore them. I know the old files have underscores, so maybe a good day to stop using them from now on in favor of dashes.

RonaldDijkstra:
Probably can make the PDF a bit smaller with export settings in whatever program you used, because I think over 1mb is a bit much for a 2 pages PDF without bitmaps

RonaldDijkstra:
Cool

source/localizable/migratie.html.erb
```diff
@@ -16,11 +16,16 @@ index: false
...
+<section class="migratie">
   <div class="row">
-    <article>
-      <%= partial "migration/content" %>
-    </article>
+      <div class="col">
+        <article>
+          <%= partial "migration/content" %>
+        </article>
+      </div>
```
RonaldDijkstra: Indentation is off here for `col`

source/stylesheets/pages/_migration.scss
```diff
@@ -11,4 +11,34 @@ body.migratie {
       color: $dark-gray;
     }
   }
+
+  :not(.footer):not(.cta-bottom) {
```
RonaldDijkstra: Bwoah... we should make this way easier (not in this PR)

source/localizable/migratie.html.erb
```diff
@@ -16,11 +16,16 @@ index: false
   </div>
 </div>
 
-<section>
+<section class="migratie">
```
RonaldDijkstra: Where's this class for?

source/localizable/migratie.html.erb
```diff
@@ -16,11 +16,16 @@ index: false
...
   <div class="row">
-    <article>
-      <%= partial "migration/content" %>
-    </article>
+      <div class="col">
+        <article>
+          <%= partial "migration/content" %>
+        </article>
+      </div>
+    <div class="col">
```
RonaldDijkstra: The alignment of the image is off, because it's a 'portrait mode image' left aligned and you use 2 columns here with the same with.

source/stylesheets/pages/_migration.scss
michellusthof: I agree. I was only able to fix the styling issues on footer and cta with help from @sn3p, fortunately.

source/localizable/migratie.html.erb
michellusthof: Yep, deprecated. Will be removed in next commit.

source/stylesheets/pages/_migration.scss
RonaldDijkstra: Thank you @sn3p 😁

RonaldDijkstra: #1708

RonaldDijkstra:
@michellusthof when you've resolved comments on a pull request please click `resolve` buttons accordingly

Issue #1721, opened 2019-03-08
Add testimonial FEK


isabelamy:
I saw that we have a quote in the FEK case. That quote should also be under 'Kundenstimmen' and on the homepage instead of Anne Monier's quote.
<img width="331" alt="screenshot 2019-03-08 at 15 32 02" src="https://user-images.githubusercontent.com/5277505/54035041-b223f680-41b8-11e9-850a-d4f4b3e6b86d.png">

RonaldDijkstra:
@lennartdefacto this is yours

RonaldDijkstra:
When? 🍔

lennartdefacto:
Working on it at this very moment.

PR #1723, opened 2019-03-11
Fix cases quote

RonaldDijkstra:
Noticed here: #1721 the quote in the sidebar is a bit off. This fixes that.


RonaldDijkstra:
https://github.com/DefactoSoftware/website/commit/ce65d7801152b8460de12062af83b1e6cc237c5d#diff-57914a4a5d1f8542422727f1b731be45 should use straight quotation marks in frontmatter @lennartdefacto

PR #1751, opened 2019-03-20
Use better argument for when we need no open graph

### LINKED ISSUE #1748: Open graph partial argument
`<%= partial "partials/meta/open_graph" unless current_page.data.robots && current_page.data.robots.include?("noindex") -%> `

When frontmatter data robots includes noindex it doesn't include the open_graph partial. 
But we shouldnt do that, because `og:image` is great for when people share a page and noindex is a wrong argument.

Probably best to use another another argument instead if we do not want open_graph to be included, probably:

`no_open_graph: true` 

Because defaults false.

RonaldDijkstra:
Close #1748


source/layouts/layout.erb
```diff
@@ -26,7 +26,7 @@
 
     <%= alternate_link_tags(current_page) -%>
 
-    <%= partial "partials/meta/open_graph" unless current_page.data.robots && current_page.data.robots.include?("noindex") -%>
+    <%= partial "partials/meta/open_graph" unless current_page.data.no_open_graph -%>
```
fatboypunk: Be aware that you're using a double negative here and you'd rather not have that

RonaldDijkstra: Thanks

PR #1770, opened 2019-03-29
Fix form width in Compliance API landing page

michellusthof:
Matching form width of new Quizzes landing page.


source/localizable/call-to-actions/_information-compliance-api.nl.erb
```diff
@@ -9,6 +9,7 @@
         Download ons handige productsheet met alle informatie.
       </p>
     </div>
+  </div>
 
     <form class="form" accept-charset="UTF-8" action="<%= t("urls.contact") %>" method="POST">
```
sn3p: form needs an indent to the left ;)

PR #1799, opened 2019-04-08
Add quick fix to use target blank on news item

RonaldDijkstra:


RonaldDijkstra:
You should've wait on deploy preview for this one :-)

PR #1801, opened 2019-04-08
Remove Bowser dependency

RonaldDijkstra:
See if this works out as expected. 

#1796


RonaldDijkstra:
This removes unused gem bowser.

PR #1802, opened 2019-04-08
Use ruby 2.4.3 in circleci setup

RonaldDijkstra:
This makes sure we use same Ruby version on Circle as locally and anywhere else (read: Netlify) 

Maybe we can introduce a env var for this.


RonaldDijkstra:
Haha i merged this myself 

image

PR #1809, opened 2019-04-09
Introducing font weight vars

RonaldDijkstra:


EasterPeanut:
Hm, we did that before and removed it later, don't know why exactly. @sn3p ?

sn3p:
> Hm, we did that before and removed it later, don't know why exactly. @sn3p ?

CSS already gives us a preset of variables (normal, bold, ..., 400, 700). To me this feels like we're aliasing aliases.

RonaldDijkstra:
We did not do this before on the websites. If you guys think this is unneeded I'll close this.

sn3p:
> We did not do this before on the websites. If you guys think this is unneeded I'll close this.

@EasterPeanut meant in detroit I think, we had the same discussion there I remember.
I try to think of variables as mappings between value and meaning. This feels a bit like mapping values to values (not bad per se of course). But I'm not bothered, it's up to you :)

RonaldDijkstra:
Yes figured she meant Detroit haha 👍

EasterPeanut:
> CSS already gives us a preset of variables (normal, bold, ..., 400, 700). To me this feels like we're aliasing aliases.

I understand this point of view, but I also understand it could add a bit of restriction if you only use the defined variables. Not that big a deal, both ways should be fine, I guess.

sn3p:
> I understand this point of view, but I also understand it could add a bit of restriction if you only use the defined variables. Not that big a deal, both ways should be fine, I guess.

I see your point making it a bit more strict and mapping available sizes to variables. Merge!

PR #1811, opened 2019-04-12
Update the description, content and download of the Update page

lennartdefacto:
- [x] Update content to use 'je' in stead of 'U'
- [x] Update PDF (link to, and the file itself)
- [x] Update the url-name to Upgrade
- [x] Redirect old URL to new URL
- [x] Delete old PDF


source/localizable/migration/_content.nl.md
```diff
@@ -1,7 +1,7 @@
-## Uw update naar CAPP LMS staat klaar
+## Uw upgrade naar CAPP LMS staat klaar
```
RonaldDijkstra: Uw = je

source/localizable/upgrade.html.erb
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Uw update naar CAPP LMS"
+title: "Uw upgrade naar CAPP LMS"
```
RonaldDijkstra: Uw = je

RonaldDijkstra: Or jouw

lennartdefacto:
@RonaldDijkstra could you review this also since I made some change after your initial review? Thanks.

config.rb
```diff
@@ -99,6 +99,8 @@
            "e-learning.html"
   redirect "learningspaces.html", to:
            "capp-agile-learning.html"
+  redirect "update.html", to:
```
RonaldDijkstra: We probably can use this from now on: 

https://www.netlify.com/docs/redirects/ 

I think those are server side instead of client side, which is slightly better SEO wise etc.

lennartdefacto: So do you want to fix this here/now, or publish this as-is and change it throughout the website afterwards through #1813?

RonaldDijkstra: After, because this impacts way more pages than just this one.

PR #1821, opened 2019-04-24
Update UPGRADE formkeep endpoint and change order of form fields

RonaldDijkstra:
By request of sales.


source/localizable/call-to-actions/_information-migration.erb
```diff
@@ -9,10 +9,10 @@
       Download ons product sheet (met handige activiteitenlijst) met alles wat je moet weten over de upgrade.
     </p>
 
-    <form class="form-inline" accept-charset="UTF-8" action="<%= t("urls.contact") %>" method="POST">
+    <form class="form-inline" accept-charset="UTF-8" action="<%= t("urls.upgrade_productsheet") %>" method="POST">
       <input type="hidden" name="utf8" value="✓">
-      <input type="hidden" name="form" value="Contact - Migratie">
       <input type="text" name="organisation" placeholder="<%= t("form.organisation") %> *" required>
+      <input type="hidden" name="form" value="Contact - Migratie">
```
michellusthof: For the purpose of avoiding all confusion, I would change the value of this input field to 'Download - Upgradesheet' since it is also reflected in the FormKeep emails.

RonaldDijkstra: Mkay

PR #1841, opened 2019-05-02
Put head stuff in a shared partial 

RonaldDijkstra:
and pass some locals to it from questionnaire layout 

No more duplicate layout.


RonaldDijkstra:
Ahh rebase

PR #1859, opened 2019-05-10
Change Dutch button captions in home hero

michellusthof:


RonaldDijkstra:
This is ready right?

michellusthof:
> This is ready right?

It is. Can you check the if loop?

PR #1871, opened 2019-05-16
Use Netlify 301s

### LINKED ISSUE #1813: Use netlify redirects (301)
https://www.netlify.com/docs/redirects/

RonaldDijkstra:
I always thought using client side redirects for changing url's was suboptimal. Because we use Netlify now we can use server side 301 rewrite rules (permanent redirects). This PR introduces just that. It writes a _redirects file to the build folder after build with the correct redirect formatting for Netlify: 

https://www.netlify.com/docs/redirects/ 

Is there a way to do this more clean? 

Close #1813


RonaldDijkstra:
Netlify reports back per deploy how many redirect rules where processed: 

NL: 
Screenshot 2019-05-16 at 17 27 00

RonaldDijkstra:
We have to work a bit more clean now because there won't be any builds of non-existing pages as was the case with the client side redirects. 

Last commits in this PR are all about updating links to pages that do not exist anymore. We always should use the proper URL in internal links.

PR #1875, opened 2019-05-17
Move partials to folder outside localizable

RonaldDijkstra:


RonaldDijkstra:
I want localizable to contain only page templates and content

RonaldDijkstra:
So call-to-actions and header and footer are gone move as well.

PR #1876, opened 2019-05-20
Build navigation from data

RonaldDijkstra:


source/partials/header/_nav-primary.erb
```diff
@@ -0,0 +1,17 @@
+<li class="nav-section primary">
+  <% primary_nav_items.each do |primary_nav_item| %>
+    <%= nav_link_to "#{primary_nav_item.title}", "#{primary_nav_item.path}", class: "root-link item-#{primary_nav_item.item} has-dropdown", "data-dropdown": "#{primary_nav_item.item}" %>
```
fatboypunk: title path and item are all strings aren't they?

source/localizable/header/_nav-primary.erb
```diff
@@ -1,15 +1,12 @@
...
+<%
+all_nav_items = data["navigation_" + I18n.locale.to_s]
+nav_items = all_nav_items.clone
+%>
 
-  <% if root_locale == :de %>
-    <%= nav_link_to t("header.solutions"), "/oplossingen.html", class: "root-link item-solutions has-dropdown", "data-dropdown": "solutions" %>
+<li class="nav-section primary">
+  <% nav_items.primary.each do |primary_nav_item| %>
+    <%= nav_link_to "#{primary_nav_item.title}", "#{primary_nav_item.path}", class: "root-link item-#{primary_nav_item.class}", "data-dropdown": "#{primary_nav_item.class}" %>
```
fatboypunk: Same here, don't think you need the interpolation

RonaldDijkstra: Oh thanks

RonaldDijkstra: True

PR #1885, opened 2019-05-24
Use flex 1 1 auto for smaller screens

RonaldDijkstra:
Well lets see if this is the correct solution for all browsers.


RonaldDijkstra:
Seems to be ok.

PR #1903, opened 2019-06-12
Use contact form partial on contact page

RonaldDijkstra:


source/localizable/capp-agile-learning.html.erb
```diff
@@ -30,7 +30,7 @@ nav_white: true
       <%= partial "partials/form", locals: { title: "Demo - CAPP Agile Learning",
                                              fields: ["name", "email", "organisation", "phone"],
                                              button_text: t("form.submit"),
-                                             variant: "contact",
+                                             variant: "",
```
michellusthof: This should be left empty? It's either 'contact' or blank?

RonaldDijkstra: Should be removed thank you

PR #1907, opened 2019-06-14
Use call-to-action partial in designated folders for posts

RonaldDijkstra:
Slowly making sure we can combine certain conversions into one goal.


source/blog/de/2018-11-22-kundentag-november-2018.html.markdown
```diff
@@ -7,7 +7,6 @@ lang: de
 image: images/blog/social/kundentag-social-1200x630.jpg
 featured_image: images/blog/featured/20181122-kundentag.jpg
 header_image: moods/kundentag-mood.jpg
-call_to_action: subscribe-newsletter
```
RonaldDijkstra: This is default so remove

source/cases/de/evangelisches-krankenhaus-oldenburg.html.markdown
```diff
@@ -23,7 +23,6 @@ facts:
   url: https://www.evangelischeskrankenhaus.de/
   since: 2015
   employees: 1200
-tags: "healthcare"
```
RonaldDijkstra: Tags have no use in cases.

PR #1932, opened 2019-07-01
Fit the long german titles in the mobile navigation

RonaldDijkstra:


RonaldDijkstra:
This is been bypassed by the work done in #1944

PR #1933, opened 2019-07-01
Clearer language switch

### LINKED ISSUE #1750: Unclear language switch
Screenshot 2019-03-19 at 16 06 10

It isnt very clear that under NL/DE/EN there's a language switch

RonaldDijkstra:
Close #1750


locales/de.yml
```diff
@@ -323,6 +323,7 @@ de:
     themes: "Themen"
     events: "Veranstaltungskalender"
     open-source: "Open Source"
+    select-language: "Wähle deine Sprache"
```
michellusthof: I would use "Ihre" instead of "deine". German language is more formal, especially when you're not acquainted (yet).

RonaldDijkstra: Ist mir egal

PR #1955, opened 2019-07-12
Add updated My List pdf

michellusthof:


RonaldDijkstra:
I don't like the gray background in the first image with all the learnables.

PR #1965, opened 2019-07-29
360: Use partial for hero and add a partial for button blocks

RonaldDijkstra:
Inspired by your Legobox @EasterPeanut here's a partial for a button block.


source/localizable/360-feedback.html.erb
```diff
@@ -1,4 +1,5 @@
 ---
+layout: "product_layout"
```
RonaldDijkstra: For da future

PR #1983, opened 2019-08-06
Fix hero image on ios

RonaldDijkstra:


RonaldDijkstra:
@michellusthof deploy preview should be tested on a real iOS device.

michellusthof:
Looks good on my device (iPhone X). Perhaps good to check on mobile devices using smaller screens as well?

PR #2003, opened 2019-08-14
Update News section

### LINKED ISSUE #1734: Design of news items
So I was thinking that the news items on the homepage are way too big. I mean it's supposed to be short news about latest trends/events/activities etc. Now it takes up the whole page.

Here some ideas:
1. I think it would be nice if you could see news and products on one screen when you scroll down.
2. Maybe news should just be a tile with a title.
3. The tiles should be smaller. Like CAPP LMS or CAPP Agile tile size.
4. Titles shouldn't be longer than two lines.

What do you guys think?

RonaldDijkstra:
Move it lower on the page 
Fix how it gets displayed (make it smaller)
Add button to more

Close #1734


RonaldDijkstra:
Screenshot 2019-08-14 at 12 02 28

PR #2006, opened 2019-08-14
Add SaaS Whitepaper

RonaldDijkstra:
Add PDF
Update Security page to make this available as a download.


locales/nl.yml
```diff
@@ -256,6 +256,9 @@ nl:
         title: CAPP LMS
   hosting:
     heading: Hosting & Security
+    cta_bottom:
+      text: Laat uw gegevens achter en download onze whitepaper met daarin alle informatie.
```
michellusthof: I think it is 'ons' whitepaper instead of 'onze'. Sounds better to me.

RonaldDijkstra: Ok!

PR #2008, opened 2019-08-19
Remove old events and linked pages from German event calendar

michellusthof:


michellusthof:
> Is the webinar.jpg background image still in use?

Don't think so, therefore removed.

PR #2010, opened 2019-08-20
Exclude Normalize.scss from linter

RonaldDijkstra:


RonaldDijkstra:
Cool. Works.

source/stylesheets/_normalize.scss
```diff
@@ -403,3 +403,7 @@ table {
     border-collapse: collapse;
     border-spacing: 0;
 }
+
+/**
+* Test for linter ignore
```
sn3p: This can be removed right?

RonaldDijkstra: Yeah sure was testing only ;-)

PR #2016, opened 2019-08-29
Add client reference for LZR

michellusthof:
Add LZR logo


RonaldDijkstra:
You don't to do those all in 1 PR.

PR #2034, opened 2019-09-16
Fix the navigations

RonaldDijkstra:
Main nav should consist of: 

Products
Services
Clients
Defacto
Blog
Vacatures 

But blog and vacatures should render in the Defacto group on mobile and in the footer, so introducing 'not_main' for these two items that should not render there.


RonaldDijkstra:
@sn3p can you review please?

RonaldDijkstra:
https://deploy-preview-2034--defacto-nl.netlify.com/

data/navigations_de.yml
```diff
@@ -93,6 +94,11 @@
     - title: Jobs
       url: /jobs/index.html
       icon: icons/jobs.svg
+      not_main: true
```
sn3p: I would go for:

```suggestion
      main: false
```

So defaults to true unless set the false.

source/partials/header/_nav-dropdowns.erb
```diff
@@ -43,11 +43,13 @@
                 <div class="link-group">
 
                   <% item.links.each do |link| %>
-                    <% nav_link_to(link.url, class: "link-container with-icon") do %>
-                      <span class="link-title link-icon">
-                        <%= inline_svg link.icon, title: link.title %>
-                        <%= link.title %>
-                      </span>
+                    <% if link.not_main == nil %>
```
sn3p: Then you can do:

```suggestion
                    <% unless link.main == false %>
```

RonaldDijkstra: Cool. Thanks. It's fixed 👍

PR #2047, opened 2019-11-06
Fix some bad aligned call-to-actions

RonaldDijkstra:
Some quick & dirty fixes for some call-to-actions on certain pages. See files changed for which pages.


source/stylesheets/pages/_upgrade.scss
```diff
@@ -12,7 +12,7 @@ body.upgrade {
     }
   }
 
-  :not(.footer):not(.cta-bottom) {
+  .content {
```
RonaldDijkstra: Not and not is not needed. Just target a specifc class and all others are not taken into account.

PR #2048, opened 2019-11-06
Add new blog about new live customers

michellusthof:
- [x] Add new social image
- [x] Add new featured image


source/localizable/call-to-actions/contact.erb
```diff
@@ -8,7 +8,7 @@
         <%= t("blog.cta_bottom.capp-lms.text") %>
       </p>
     </div>
-    <%= partial "partials/form", locals: { title: "Contact Request: migratie-update blog nov 2019",
+    <%= partial "partials/form", locals: { title: "Contact Request: Bllog - MUMC en Spaarne live nov 2019",
```
fatboypunk: Is the `Bllog` intentional?

michellusthof: Eh, no 😅

PR #2077, opened 2019-12-18
Update news items on homepage

michellusthof:


RonaldDijkstra:
Can we make em all the same size?

PR #2082, opened 2020-01-15
Update CAPP LMS product page to CAPP Learning proposition

michellusthof:
- [ ] Add new icons for USPs
- [ ] Add new testimonials
- [ ] Add new illustrations for product overview (@fritsypeij)
- [ ] Update client selection  
- [ ] **Add redirect when changing path to /capp-learning**


michellusthof:
For consideration: Use the new URL /capp-learning or retain /capp-lms because of its history in search engines and our brand value?

RonaldDijkstra:
Once this page had just /capp as the path. For SEO purposes I added LMS, because using keywords in the URL is a very solid SEO technique. At that time I wanted us to be (better) found on LMS, because on that and related keywords is where our volume is when talking traffic wise. 

But I like the 'branding as a whole' approach too. 

At least, when you do use a new path, do not forget to add the redirect to the specific data file to make sure Netlify returns a 301. When done properly we will not lose that many Link Juice.

PR #2091, opened 2020-01-23
Add video to hero of CAPP Compliance

RonaldDijkstra:
Because:
- Video is ready to be published somewhere in the next coming days

This commit:
- Adds a designated hero with the video embed
- Refactors the stylesheet for the page a bit

Thoughts:
- We should definately get rid of the product_layout as
  that causes abstractions where we must overrule some things
  we should just write specific styling per product page
  as we used to 

Todos: 
- [ ] Add real video ID (should be on Defacto account)


source/stylesheets/base/_grid-settings.scss
```diff
@@ -2,21 +2,45 @@ $max-width: 1260px;
...
 
 // Menu media queries
-$menu-mobile: (media: "screen and (max-width: 999px)");
-$menu-desktop: (media: "screen and(min-width: 1000px)");
+$menu-mobile: (
+  media: "screen and (max-width: 999px)"
+);
+$menu-desktop: (
+  media: "screen and(min-width: 1000px)"
+);
```
RonaldDijkstra: Autoformatted...

PR #2106, opened 2020-02-10
Fit the main navigation with more items by using flexbox

RonaldDijkstra:
This makes the main navigation a bit more flexible with flex box.


RonaldDijkstra:
Tested in IE10, IE11, Edge, Chrome, Safari and Firefox. All good

PR #2118, opened 2020-02-21
Fix Testimonials flexbox x IE11

RonaldDijkstra:
afbeelding

Some other not so good things with testimonials get fixed too here.


source/stylesheets/config/_variables.scss
```diff
@@ -2,7 +2,7 @@
 $header-height: 74px;
 
 // Typography
-$base-font-family: 'Open Sans', $font-stack-helvetica;
+$base-font-family: "Open Sans", $font-stack-helvetica;
```
RonaldDijkstra: Prefer double over single quotes here.

PR #2148, opened 2020-04-02
Add new blog about CAPP LMS corona stats

michellusthof:


RonaldDijkstra:
I think the /path can use some love. For example this is a bit more describing the content of the post: 

`/corona-crisis-onderstreept-belang-digitaal-leren`

Other than that: nice post. Good job.

PR #2154, opened 2020-04-10
Remove Dutch e-learning and redirect to new solution page

michellusthof:


data/redirects_nl.yml
```diff
@@ -16,3 +16,4 @@
 - /compliance-api /capp-compliance-api
 - /capp-lms /capp-learning
 - /hosting-security /saas
+- /e-learning /maak-zelf-e-learning
```
RonaldDijkstra, line 19: We should abc the redirects data. Not for this PR though

RonaldDijkstra:
Are we sure there are no links left in blogs?

michellusthof:
Already checked, but I haven't found any forgotten links.

PR #2166, opened 2020-04-22
A partnerships page

RonaldDijkstra:
Adds a page about partnerships and stuff in the about us navigation.


RonaldDijkstra:
@michellusthof do we continu this one?

PR #2179, opened 2020-05-14
Target Firefox for Open Sans weight not rendering properly

RonaldDijkstra:
Since one of the latest updates, headings on the site do not render properly. Although not set, all of them with the $font-weight-small (aka 300) render in _italics_. 

This branch targets this and sets a different font-weight for Firefox only (aka 400). At least that looks better than all the headings in italic.


source/stylesheets/base/_typography.scss
```diff
@@ -65,11 +70,11 @@ a {
 
 // Using what-intent to target specific input
 // https://github.com/ten1seven/what-input
-[data-whatintent='mouse'] *:focus {
+[data-whatintent="mouse"] *:focus {
```
RonaldDijkstra, line 73: autoformatting

RonaldDijkstra:
For now I don't mind the website looks slightly different in Firefox than in Chrome. Hard to say what the problem really is here.

PR #2183, opened 2020-05-18
Separte the summary on </p>

fatboypunk:
Before the summery on the index blog page would include the first h2 at
times. This would be concated right at the end of the summary giving a
weird text. With this seperator the summary will only take the first
paragraph to show the summary and nothing more. This is how it mostly
feels that should happen while writing blogs.

before:
Screenshot 2020-05-18 at 15 37 04
With  this update 
Screenshot 2020-05-18 at 15 37 00


RonaldDijkstra:
Nice

michellusthof:
Cool!

PR #2189, opened 2020-05-20
Ignore 503 errors in HTMLProofer

michellusthof:
Considering the many false positive 503 errors generated by HTMLProofer lately, we might want to ignore these (temporarily).


RonaldDijkstra:
I understand the irritation, but I think globally ignoring 503's is not the way you want to do this.

PR #2191, opened 2020-05-25
Make some changes to the klantendag form

michellusthof:
Some form fields were required where they shouldn't be. Other fields didn't indicate whether they were required or not. This PR resolves both.


michellusthof:
[Preview](https://deploy-preview-2191--defacto-nl.netlify.app/klantendag/)

PR #2208, opened 2020-07-02
Remove back link from thanks page

RonaldDijkstra:
Not in all scenarios there's a back link because in some cases the user gets to the page from a click on a link in an email. 
And for regular filled in forms it's a bit strange to go back to the form after just submitting it. 

So I decided that it should be removed.


michellusthof:
I agree with this decision (fwiw) 😛

RonaldDijkstra:
HTML proofer outputs 429 errors for twitter.com links. I don't even know 429.

RonaldDijkstra:
Too many requests ok...

